PR #103, opened 2019-11-17
Add Python based Allied Vision code

neildhar:
Adapt pymba python library to do continuous imaging and publishing with ROS. See README for running instructions


src/avt_camera/scripts/mono_camera.py
```diff
@@ -0,0 +1,87 @@
...
+                rospy.logerr("No cameras found, big RIP")
+                sys.exit(0)
+
+            for cam_id in camera_ids:
+                rospy.loginfo("Camera found: " + cam_id)
+
+            if  self._camera_id in camera_ids :
+                c0 = vimba.getCamera(self._camera_id)
+            elif self._camera_id is None :
+                c0 = vimba.getCamera(camera_ids[0])
```
davidAmiron: Should check if cameras were found

src/avt_camera/scripts/mono_camera.py
```diff
@@ -0,0 +1,87 @@
...
+        with Vimba() as vimba:
+            system = vimba.getSystem()
+            system.runFeatureCommand("GeVDiscoveryAllOnce")
+            rospy.sleep(0.1)
+
+            camera_ids = vimba.getCameraIds()
+            if not camera_ids:
+                rospy.logerr("No cameras found, big RIP")
+                sys.exit(0)
+
```
davidAmiron: Somehow check if the same camera is requested and give a reasonable error message if that happened

src/avt_camera/scripts/mono_camera.py
```diff
@@ -0,0 +1,87 @@
...
+        
+        with Vimba() as vimba:
+            system = vimba.getSystem()
+            system.runFeatureCommand("GeVDiscoveryAllOnce")
+            rospy.sleep(0.1)
+
+            camera_ids = vimba.getCameraIds()
+            if not camera_ids:
+                rospy.logerr("No cameras found, big RIP")
+                sys.exit(0)
```
davidAmiron: Error, exit with non zero exit code

src/avt_camera/scripts/mono_camera.py
```diff
@@ -0,0 +1,87 @@
...
+            for cam_id in camera_ids:
+                rospy.loginfo("Camera found: " + cam_id)
+
+            if  self._camera_id in camera_ids :
+                c0 = vimba.getCamera(self._camera_id)
+            elif self._camera_id is None :
+                c0 = vimba.getCamera(camera_ids[0])
+            else:
+                rospy.logerr("Requested camera not found, sorry")
+                sys.exit(0)
```
davidAmiron: Error, exit with non zero exit code

src/avt_camera/scripts/mono_camera.py
```diff
@@ -0,0 +1,87 @@
...
+                rospy.loginfo("Packet Size: " + str(c0.GevSCPSPacketSize))
+                rospy.loginfo("Stream Bytes Per Second: " + str(c0.StreamBytesPerSecond))
+                c0.runFeatureCommand("GVSPAdjustPacketSize")
+                c0.StreamBytesPerSecond = 100000000
+            except:
+                #not a gigE camera
+                pass
+
+            #set pixel format
+            c0.PixelFormat="Mono8"
```
davidAmiron: Where is documentation for these parameters?

src/avt_camera/scripts/test.py
```diff
@@ -0,0 +1,75 @@
+# -*- coding: utf-8 -*-
```
davidAmiron: Remove this file

davidAmiron (dismissed):
Remove test.py

src/avt_camera/scripts/mono_camera.py
davidAmiron: Should already do that

src/avt_camera/scripts/mono_camera.py
davidAmiron: never mind

src/avt_camera/scripts/mono_camera.py
davidAmiron: Look at vimba viewer, ctrl-f in pymba

src/avt_camera/scripts/mono_camera.py
davidAmiron: neil: "try catch on opencamera"

src/avt_camera/scripts/stereo_camera.py
```diff
@@ -0,0 +1,147 @@
...
+			rframe.announceFrame()
+
+			c0.startCapture()
+			c1.startCapture()
+
+			framecount = 0
+			droppedframes = []
+            
+			c0.runFeatureCommand("AcquisitionStart") 
+			c1.runFeatureCommand("AcquisitionStart")
```
neildhar: there's a lot of code duplication here, consider using a function or a loop to do the setup on both cameras simultaneously

davidAmiron:
This is temporarily in backlog, we are getting the jetson working with GPUs until we can test this some more.

src/avt_camera/scripts/stereo_camera.py
muthuArivoli: Most, if not all duplication was removed in 3bec872. Last bit (between monocamera and synchronouscamera) can be figured out later, if we decide that synchronous camera is good enough without mono camera.

muthuArivoli:
Things to do before merge
- [ ] Test single camera, especially whether output has been formatted appropriately for CV
- [ ] Test synchronized camera
- [ ] Test throwing exceptions 
- [ ] Add calibration for downward camera
- [ ] Add launch file for downward camera

Notes:
This new code will need to be tested thoroughly. It should be functionally equivalent to the code from the last time we tested except for two things:

1. An exception has been removed in the vimba_frame header file, specifically the `waitFrameCapture` function. Previously, we had the pymba library in this package, and had manually added that exception in. Since the library has been moved to the docker image, we don't have access to the library files, so it currently will not throw an exception. If this causes problems, then we can process the exception in our code itself, by changing `self._frame.waitFrameCapture()` to 
```
err = self._frame.waitFrameCapture()
if err != 0:
    raise VimbaException(err)
```

2. The code should now be able to detect if a camera is requested a second time, and will exit with error. Not completely sure if this will work, will definitely need to test for false positives on restarting a camera. If there are indeed false positives, then the only other way I can currently see is to get the private variable  `_cameras` from our Vimba instance. Using this method, we would check if the camera_id is in the keys already, immediately before `self._c0 = vimba.getCamera(self._camera_id)`.  
```
if self._camera_id in vimba._cameras:
    sys.exit(1)
```
This is definitely not good coding style, accessing a private variable like this, but I was not able to find any other way.

onboard/catkin_ws/src/avt_camera/calibrations/DEV_000F315C1ED5.yaml
```diff
@@ -0,0 +1,18 @@
+!!python/unicode 'camera_matrix':
```
davidAmiron, line 1: How were these generated? Do they work?

onboard/catkin_ws/src/avt_camera/calibrations/DEV_000F315C1ED5.yaml
```diff
@@ -0,0 +1,18 @@
...
+  !!python/unicode 'data': !!python/tuple [1333.7083757875523, 0.0, 961.7952043539852,
+    0.0, 1342.2102793367508, 600.5856686666136, 0.0, 0.0, 1.0]
+!!python/unicode 'camera_name': DEV_000F315C1ED5
+!!python/unicode 'distortion_coefficients':
+  !!python/unicode 'data': !!python/tuple [-0.1649891525322297, 0.09538162440134604,
+    -0.00405605433998157, -0.005871950362120829, 0.0]
+!!python/unicode 'distortion_model': plumb_bob
+!!python/unicode 'image_height': 760
```
davidAmiron, line 9: They are yaml files, can we remove the `!!python/unicode` from all the lines?

onboard/catkin_ws/src/avt_camera/scripts/synchronous_camera.py
```diff
@@ -0,0 +1,48 @@
+#!/usr/bin/env python
+
+from pymba import *
+import rospy
+from sensor_msgs.msg import Image, CameraInfo
+from camera import Camera
+
+
+class SynchronousCamera:
```
davidAmiron, line 9: Rename to SynchronizedCameras

onboard/catkin_ws/src/avt_camera/calibrations/DEV_000F315C1ED5.yaml
muthuArivoli: They do work. I believe they were generated using the camera calibration package in ROS.

muthuArivoli:
This has been tested with both the left and right camera individually, and with them together for stereo vision. All the exceptions have been tested like trying to call a camera twice and trying to run without a camera attached. This should be good to merge.

ShaanGondalia (approved):
LGTM

rabinowitz (changes requested):
LGTM! One addition: please update the list of packages in the repo's main README to include this new one.

rabinowitz (approved):
LGTM! Squash and merge when ready.

muthuArivoli (approved):
This is finally landing today

Issue #83, opened 2019-10-19
Write Task superclass


davidAmiron:
Write the class task that will be overrided by all other tasks

neildhar:
I'd like to propose a change to the task architecture. There are a couple issues I identified with the current one:

- Difficulty of creating subtasks
- Constantly creating and destroying subscribers and publishers for every task is confusing
- We need simple, easy to construct tasks to make the program easy to reason about

The current task architecture is pretty complicated. Every task makes its own subscriber and publisher (which might even fail if the previous publisher was not correctly destroyed). They also store a lot of state and generally share a lot of functionality.  I think there are a couple of changes we should make:

## Separate utility functions
 Pull out the utility functions (such as distance calculation, moving, angles, and pose transforms) into a separate module called TaskUtils. These functions would take at most an additional Pose parameter and they would be easier to generalise and reason about.

## Consolidate shared state
Somehow make the state that is shared across tasks global (this includes the current location and the handles for the publisher and subscriber). This could mean that the top level module (currently task_planner.py) stores all this state and passes a handle to each task when it is initialised. Ideally, we'd just make these static members of TaskPlanner but I don't know if you can do something like that in Python.

## Change TaskBase
With these changes, the tasks should become very lightweight, with almost no state (apart from pose at task start) and only the `init` and `run` methods. The `init` function in TaskBase should now only need to initialise what little state there is and would almost never need overriding.

## Implement template tasks

This is probably the most interesting part. We create a series of derived classes that just need a constructor argument at instantiation to become a usable task. Off the top of my head, I can think of the following tasks:

- **MoveToGlobal:** Given desired global pose as a constructor argument (relative to mission start), end the task at that pose.
- **MoveToLocal:** Same but with a local pose relative to task start
- **TimedMove:** Move in a given direction for a given amount of time
- **FollowState:** Given a callback function or ROS topic corresponding to some processed sensor data (say, the target position/velocity desired to hit a buoy), set the robot's desired state to correspond to that. This is useful for generalising to a lot of tasks like the pinger, buoy, etc and provides a standardised interface for the people developing that code.
- **ListTask:** Given a list of tasks, initialise and execute each one in order. This is useful for composing tasks and creating subtasks. Having these will make it easier to organise tasks into logical units and avoid making tasks too big.

[...]

PR #144, opened 2020-01-26
Add ListTask, simplify dependencies, fix bugs

neildhar:
Extract state from the tasks and place it in a TaskState object. This avoids constantly creating and destroying publishers and subscribers.

Create a ListTask that just executes a series of tasks in order. See #83.


src/motion_planning/scripts/task.py
```diff
@@ -2,11 +2,11 @@
...
 from geometry_msgs.msg import Pose, Twist
 
-from abc import ABC, abstractmethod
+from abc import ABCMeta, abstractmethod
 
-class Task(ABC):
+class Task:
     """High level task that represents some function"""
-
+    __metaclass__ = ABCMeta
```
davidAmiron: Why do you choose to do it this way instead of inheritance from ABC?

src/motion_planning/scripts/task_runner.py
```diff
@@ -1,44 +1,23 @@
...
+
 import rospy
 from task import Task
 from template_tasks import *
-from competition_task import CompetitionTask
 
 class TaskRunner(object):
+    RATE = 30 # Hz
 
-	RATE = 30 # Hz
```
davidAmiron: Why do you take all of the Task State and put it in another class? This makes it more difficult to trace and I do not see what it adds.

src/motion_planning/scripts/task_state.py
```diff
@@ -0,0 +1,29 @@
...
+import rospy
+
+class TaskState(object):
+    STATE_TOPIC = 'state'
+    DESIRED_POSE_TOPIC = 'controls/desired_pose_global'
+    DESIRED_TWIST_GLOBAL_TOPIC = 'controls/desired_twist_global'
+    DESIRED_TWIST_LOCAL_TOPIC = 'controls/desired_twist_local'
+
+    # Initialise static members that will hold the publishers and listeners
+    STATE_LISTENER = None
```
davidAmiron: While it is true that there will probably only be one TaskState at a time, these are not meant to be static variables

src/motion_planning/scripts/task_utils.py
```diff
@@ -96,10 +98,10 @@ def transform(origin, destination, odometry=None, pose=None):
     return transformed 
 
 def publish_desired_pose_global(pose):
```
davidAmiron: These publishing methods should not be in task utils. Every method in task_utils should not require any state, these publishing methods are instance methods of a task.

src/motion_planning/scripts/task_utils.py
```diff
@@ -81,14 +84,21 @@ def transform(origin, destination, odometry=None, pose=None):
...
-	transformed = tf2_geometry_msgs.do_transform_pose(pose, trans)
+    elif(isinstance(odomORpose, Pose)):
+	    transformed = tf2_geometry_msgs.do_transform_pose(odomORpose, trans)
     return transformed 
+
+def publish_desired_pose_global(pose):
+    TaskState.DESIRED_POSE_GLOBAL_PUBLISHER.publish(pose)
+
+def publish_desired_twist_global(twist):
+    TaskState.DESIRED_TWIST_GLOBAL_PUBLISHER.publish(twist)
```
davidAmiron: Global variables are a very messy way of dependency injection

src/motion_planning/scripts/task.py
neildhar: You can't inherit from ABC in python 2

src/motion_planning/scripts/task_runner.py
neildhar: 1. The stuff that's currently stored as state in a task isn't really task specific. It's more the overall state of the robot. It doesn't make sense to keep duplicating this and it's more intuitive to just store this in a single place.

2. Constantly creating and destroying publishers and subscribers is expensive. They need to be registered with ROS and the unnecessary object creation creates GC pressure.

src/motion_planning/scripts/task_state.py
neildhar: Why not? They are static because this is the cleanest way to implement a singleton in python.

src/motion_planning/scripts/task_utils.py
neildhar: They don't store any state in task utils, the state is stored in TaskState. There are other methods in task utils that rely on global state (such as transform lookups).

I don't think publishing methods should be instance methods of a task, there's nothing task specific about them.

src/motion_planning/scripts/task_utils.py
neildhar: Again, I think our primary disagreement is about whether the publishers and subscribers are task specific state.

src/motion_planning/scripts/task.py
davidAmiron: Got it

PR #163, opened 2020-02-19
Refactor state republisher

davidAmiron:
Move to sensor_fusion, change publishing method


davidAmiron:
Did not create branch off of master, has bunch of junk in it.

PR #165, opened 2020-02-19
Fix thruster code

muthuArivoli:
Reverse certain thrusters to provide a uniform standard for the direction that a thruster moves in

### DIFF
--- a/catkin_ws/src/offboard_comms/Arduino Sketchbooks/ThrusterServoArduino/ThrusterServoArduino.ino
+++ b/catkin_ws/src/offboard_comms/Arduino Sketchbooks/ThrusterServoArduino/ThrusterServoArduino.ino
@@ -24,6 +24,13 @@ MultiplexedBasicESC *thrusters[NUM_THRUSTERS];
 void thruster_speeds_callback(const offboard_comms::ThrusterSpeeds &ts_msg){
     //copy the contents of the speed message to the local array
     memcpy(thruster_speeds, ts_msg.speeds, sizeof(thruster_speeds));
+    //negate the thruster speeds to match standard assumptions 
+    thruster_speeds[0]*=-1;
+    thruster_speeds[1]*=-1;
+    thruster_speeds[3]*=-1;
+    thruster_speeds[4]*=-1;
+    thruster_speeds[5]*=-1;
+    thruster_speeds[6]*=-1;
     last_cmd_ms_ts = millis();
 }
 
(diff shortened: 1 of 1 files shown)


neildhar:
I don’t think this should be done on the Arduino. The python scripts running on the computer are:
1. Easier to modify
2. Changed more often
3. Modifiable during a pool test

The Arduino code should do the bare minimum required to move the thrusters and actuators and then leave the rest to higher level control. The higher level control should also be the only place you code in knowledge of how the thrusters are oriented and connected.

We’ve had instances in the past where the code is modified on different instances and the old code base had at least 3 places where the thrusters directions were flipped.

5happy1:
I agree with your final point about keeping thruster directions from being flipped in multiple places, which is why we an extensive discussion to choose where to make that flip. While you are correct in your points of the benefits of doing so in the Python script, we don't derive much from these benefits. The only time directions should be flipped is when hardware changes, which only happens when the stack is accessible and the Arduino code can be updated. Furthermore, our goal is to make the CS has hardware-agnostic as possible, so ensuring that the robot has a standard interface in which providing a positive power means the thruster moves in its forward direction achieves hat goal.

neildhar:
The CS stack isn't currently hardware agnostic, and I'm not sure there would *really* be any benefit to making it so.  I agree that the hardware specific components should be kept separated but that's already done. For example:
1. We hard code the thruster layout and numerical mapping in the control script.
2. The PID parameters are hard coded in and encode all kinds of hardware information like the force produced by the thrusters and the design of the frame.
Adding which thrusters are flipped really isn't that different.

Updating the Arduino code on the other hand is complicated and unnecessary, especially since this flipping will end up being the only hardware or software dependent parameter. If we were putting all of controls on the Arduino perhaps that would make a little more sense (but would be a bad idea for other reasons).

Imagine being at competition and having a thruster stop working. Replacing a thruster is unpleasant enough, but you'll then have to test it and then possibly reflash the Arduino if the new thruster is wired up wrong.

5happy1:
I am convinced.

The code is *mostly* hardware agnostic, but not in those two places that you mentioned: a configuration file holds thruster information, and a launch file holds PID parameters. Adding a flipped option for the thrusters in the configuration file should not be much of an issue, and still keeps the hard-coded information separate from the generic controls algorithms.

We will revert the changes to the Arduino code so that it has no say in flipping, and we will be extremely careful to communicate to all parties involved that flipping thrusters only ever happens in the configuration file (at the moment, there aren't really any other places it *could* happen, which is already an improvement over the previous codebase.)

Likely, once we have the flipping set, we won't have to change much, but I do agree that there may a couple instances down the road, especially at competition, that this will be the better way to have it set up for quick fixes.

davidAmiron:
We did not end up using this method, and are instead keeping track in the controls package.

PR #180, opened 2020-03-03
Motion planning

echang264:
Merge before testing prequal task in pool


catkin_ws/src/motion_planning/scripts/combination_tasks.py
```diff
@@ -0,0 +1,54 @@
...
+
+	def __init__(self, tasks, *args, **kwargs):
+		super(SimulTask, self).__init__(*args, **kwargs)
+
+		self.tasks = tasks
+
+	def _on_task_run(self):
+		self.all_finished = True
+		for task in self.tasks:
+			task.run()
```
davidAmiron: I know a finished task will not run anyway, but it seems like there should still be logic here that is only running unfinished tasks

catkin_ws/src/motion_planning/scripts/transform_tester.py
```diff
@@ -0,0 +1,99 @@
+#!/usr/bin/env python
```
davidAmiron: Lets put this file in another folder somewhere called test, maybe in motion_planning

PR #182, opened 2020-03-04
Converting Global PID Outputs to Local Frame

ShaanGondalia:
This code converts our global PID outputs to the local frame so we allocate correct thruster values.

### DIFF
--- a/catkin_ws/src/controls/launch/controls.launch
+++ b/catkin_ws/src/controls/launch/controls.launch
@@ -7,6 +7,8 @@
         <arg name="sim" value="$(arg sim)" />
     </include>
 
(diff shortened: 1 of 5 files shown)


catkin_ws/src/controls/launch/controls.launch
```diff
@@ -7,6 +7,8 @@
         <arg name="sim" value="$(arg sim)" />
     </include>
 
+    <!--node pkg="tf2_ros" type="static_transform_publisher" name="base_link_publisher" args="0 0 0 -0.7070727 0 0 0.7071408 odom base_link" /-->
```
davidAmiron: Remove this when you no longer need it.

PR #184, opened 2020-03-07
Controls cleanup

ShaanGondalia:
Cleaned up code from pool test, fixed the problem of holding onto state after no desired state was published. Also, the issue of controls crashing with a power input from two directions should be fixed (arithmetic error). This branch is an extension of the joystick-test branch, so if you would like to you can merge this branch instead.


5happy1:
Everything looks good to me! Good bug squashing based on the results of the pool test, and a good start on some refactoring and cleaning up of the controls code, though there is obviously more to do in that department.

We have to remember to fix the IMU code to remove the jank code once we can get the transforms working. Also, we have to update our readme to reflect our controls changes once we've finalized more of them.

davidAmiron:
@autopep8 pep8ify

PR #189, opened 2020-05-22
modified runSim.sh for new docker container

puzzler7:


simulation/runSim.sh
```diff
@@ -1,8 +1,8 @@
-COPPELIA_SIM_SCRIPT=~/dev/coppelia/coppeliaSim.sh
+COPPELIA_SIM_SCRIPT=~/docker-build/coppelia/coppeliaSim.sh
```
muthuArivoli, line 1: Do you want this script in ~/dev/coppelia/coppeliaSim.sh? Should be an easy fix in the dockerfile if you want to move it.

simulation/runSim.sh
```diff
@@ -1,8 +1,8 @@
...
 SERVER="comm_server_scene.ttt"
 echo "Starting CoppeliaSim"
 
-SERVER_LOCATION=$(find /home/duke/dev -name $SERVER | head -1)
+SERVER_LOCATION=$(find /root -name $SERVER | head -1)
```
muthuArivoli, line 5: Do you mean `/root/dev` for this, since we still have a dev folder inside root?

puzzler7: /root/dev will also work, and would probably be better - the scene file will always be in the git repo.

simulation/runSim.sh
puzzler7: I don't have any opinion - both work. The entire coppelia folder would need to be moved, but that shouldn't be difficult.

simulation/runSim.sh
muthuArivoli: then, could you also change that as well?

PR #193, opened 2020-05-24
Update CMake and package files

muthuArivoli:
Updates CMake and package files to reflect our actual dependencies. Uses conditional builds to allow for our two docker images to use the same codebase when compiling. 

Needs DukeRobotics/robosub-ros-docker#45 for builds over ssh (bash script works fine).


catkin_ws/src/controls/CMakeLists.txt
```diff
@@ -1,203 +1,44 @@
...
-## INCLUDE_DIRS: uncomment this if your package contains header files
-## LIBRARIES: libraries you create in this project that dependent projects also need
-## CATKIN_DEPENDS: catkin_packages dependent projects also need
-## DEPENDS: system dependencies of this project that dependent projects also need
 catkin_package(
-#  INCLUDE_DIRS include
-#  LIBRARIES controls
-#  CATKIN_DEPENDS other_catkin_pkg
-#  DEPENDS system_lib
+  LIBRARIES controls
```
neildhar, line 42: I'm not familiar with how controls is structured but if it's standalone and talks over messages you shouldn't need to export it as a library.

catkin_ws/src/data_pub/CMakeLists.txt
```diff
@@ -1,199 +1,43 @@
...
-## * In this file (CMakeLists.txt):
-##   * add "message_generation" and every package in MSG_DEP_SET to
-##     find_package(catkin REQUIRED COMPONENTS ...)
-##   * add "message_runtime" and every package in MSG_DEP_SET to
-##     catkin_package(CATKIN_DEPENDS ...)
-##   * uncomment the add_*_files sections below as needed
-##     and list every .msg/.srv/.action file to be processed
-##   * uncomment the generate_messages entry below
-##   * add every package in MSG_DEP_SET to generate_messages(DEPENDENCIES ...)
+endif ("$ENV{COMPUTER_TYPE}" STREQUAL "onboard")
```
neildhar, line 24:
```suggestion
endif ()
```

muthuArivoli:
The difficulty in having the same full set of dependencies for both, is that those dependencies have to be installed, otherwise the build will fail. Take for example, offboard_comms. It requires rosserial (among other things). If we try to build that package (with the dependency of rosserial listed in find_package) then the build would fail on the landside image, since landside does not have rosserial installed. In the same way, the joystick package, with its dependency on ros-joy, would fail on the onboard computer since that package is not installed there.

Since we do not have any C++ nodes, we could just leave it at catkin being the only package required (since we are not compiling and linking anything). However, the main reason that I am cleaning this up is so that we can be notified when a required package isn't found, ideally through an autobuild format similar to the docker repo. This can also help us make sure that the necessary dependencies are installed before a test, rather than realizing something isn't installed, and trying to install it with the robot 10 ft underwater.

The main (and only) difference between the two build configurations is trying to determine whether the necessary packages have been installed in the docker image that uses them. I don't like it either, but I couldn't figure out a good way to keep dependencies and build our code on both images.

neildhar:
There are three other paths I can see, depending on what kind of work you're envisioning for the landside image.

1. If the landside image is going to be tightly integrated and need the ability to send messages and so forth, then we should just install all the dependencies in `core`. This doesn't incur any additional size overhead since the layers are shared with `onboard` anyway and saves some headache in separating the build rules for the two.
2. If there are distinct functions that some packages do that you feel need to be available on both, then perhaps it makes sense to split the package.
3. If the landside image is just going to do simulation and joystick stuff, then there's no need for any package support at all.

I think modifying the build rules to work differently on different platforms could end up getting pretty confusing/messy over time.

muthuArivoli:
1. If we add all the dependencies in the core image, wouldn't that just cause a lot of additional bloat in the landside image? The landside image is quite small to pull, but adding the onboard dependencies to core would add a lot of unneeded stuff, since when pulling you would download all of the unnecessary packages from core as well. One thing to add, is that I would also like to find a way to check if non-catkin python dependencies (like numpy or pytorch) have been installed, so not sure how that would work in combining dependencies.   

2. Sure, a package should only operate on onboard or only on landside, it should not be on both. However, the ability to send, and more importantly, read custom messages needs all of the packages to be built on landside. We should be able to rostopic echo topics with custom messages on landside, and that will require the messages that are generated for onboard to be generated for landside as well. Also, the GUI that we currently have can display all messages that have been defined, so we would still have to build all packages to access those custom messages, since we only want to run the GUI on landside only.

3. I'm not entirely sure what this means, but we would probably want to do more stuff. We are planning on making a CLI/GUI, so that might require additional stuff. Also, I think the point above about custom messages also applies here.

To me, as long as the only thing we do is identify packages, it should be understandable. For instance boilerplate would look like 
```
#Change X to computer this runs ons (landside or onboard)
if ("$ENV{COMPUTER_TYPE}" STREQUAL "X")
  find_package(catkin REQUIRED COMPONENTS
    # Add necessary dependencies
  )
else()
  find_package(catkin REQUIRED)
endif ()
```

neildhar:
1. I think it's very very rare that someone would download just the landside image without the onboard image.
2. The solution to this, and the way it is done in ROS, is to split the messages into a separate package. The messages are really common dependencies for multiple packages and it makes sense to organise them that way (rather than tying them to a specific package).

muthuArivoli:
Isn't the point of the landside image to not need to download the onboard image on one of our computers, and to only download the onboard image on the robot computer? Only if you do simulation, do you need both on the same computer, but for pool tests, you should only need them on separate computers.

Anyways, I thought about this more, and I think the problem is that I'm trying to force build dependencies, when we really only need run dependencies (since we only use python). Our system can be built on anything that has minimal ROS on it, but it needs those special packages to run. 

To this end, I was thinking that maybe I could write a small python script that goes into each package, reads the package.xml file, and checks whether the python-modules specified there exist (we could place the deps under the export tag, which is meant for other non-ROS stuff to interact with the package). 

This would solve the problem of checking for non-catkin python deps, since I couldn't find a way to do this using cmake. The catkin build system would then merely check if catkin exists, and would also generate messages for the packages that have custom msgs (i'll look into refactoring them into their own package in the next pr or so). So, the catkin build no longer depends on what image it is on, and we can attach a tag to the package.xml file noting which image this is supposed to run on, that the python file will check when it tests that package. This would fit nicely with our planned CLI that will presumably be written in python. 

This python script could be run in the same way we do catkin build, we could even use catkin to maybe install the python script when we run catkin build. This would also be easy to automate - you just run the script after or before you catkin build something.

I'm just very hesistant on making the core image even larger. If we add all dependencies to core, then the only thing that landside/onboard are doing are setting ssh up, and I don't believe thats enough to warrant an entirely new image. Rather than make people ssh into landside, we could just force them to exec /bin/bash into the container.

Let me know what you think.

muthuArivoli:
This might also work: http://wiki.ros.org/catkin/Tutorials/workspace_overlaying
Specifically the part about chaining workspaces. We could chain landside on top of onboard or something like that.

neildhar:
I'm in favour of doing as much as possible using ROS infrastructure and not creating magic python scripts that might cause confusion over time.

As long as overlaying is only used to add packages rather than override them, I think it seems like the right way forward. That would however entail pulling all our messages out into a separate package as previously mentioned.

I think overlaying also fits nicely with the way the Docker images are structured so it seems like a consistent conceptual distinction to draw. And it'll keep the build rules simple.

@davidAmiron what do you think?

davidAmiron:
This is starting to seem complicated. Like, very very complicated and I don't like that. Especially the separate build stuff. We would need to talk in person about what exactly needs to change, but I would like to keep this as simple as possible while still being totally functional. Splitting up the docker stuff into onboard and landside seemed reasonable, but what exactly is the reason that we need all this separate build stuff?

davidAmiron:
I am particularly opposed to a python script that checks dependencies, if we need to use a separate script for dependency management then it seems to me like a signal that we are overcomplicating.

build.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+# A convenience script used to build our code. Takes one argument that specifies the workspace to build.
+
+# If sourcing file then return instead of exiting shell
```
neildhar, line 4: I don't think we need this much error handling, we should just be able to rely on this script running correctly.

That said, I would not be opposed to spawning a subshell to run everything first, and then just sourcing the `setup.bash` scripts after that subshell returns.

PR #201, opened 2020-06-20
Acoustics Processing

muthuArivoli:
Creates the acoustics package and contains a script to do acoustics processing. Opening now for review.

Need to discuss whether the input to the action is too much - specifically if we need to change if_double and version. Also need to discuss if we want to get rid of the plotting stuff, since the plotting isn't useful during pool tests, it's only useful afterwards, when we aggregate the data together.

Will add README once all this is figured out.


muthuArivoli:
Closing now, will reopen once everything, like guess and sampling is in here.

PR #205, opened 2020-07-04
integrated guess

reedchen19:
added .action file, server, and python script for guessing the location of the pinger with the cheap set of hydrophones


reedchen19:
Thanks for the suggestions! I followed all of them and changed the correlation script to use list comprehension. I moved the filtering and plotting functions to acoustics_math, and changed the feedback to use uint32 curr_stage and total_stage.

PR #207, opened 2020-07-21
Google Drive package

muthuArivoli:
This provides a package to upload and download files from Google Drive.

The main use case is during pool testing, when we need to download a large file from drive (like a CV model). Currently, we have to  download it onto our computer (which could be very slow if we do not want to disconnect from the robots network and connect to DukeVisitor), and then use scp or ftp to transfer the file onto the robot, which could again be slow with a bad tether connection. This package basically bypasses our tether/router and downloads the file directly into the container, potentially saving us valuable time during pool tests.


muthuArivoli:
Probably not useful.

PR #211, opened 2020-09-11
Remove legacy task planning package and motion planning notes

muthuArivoli:
Deletes the task planning package, which I believe has been superseded by motion planning. Also deletes the motion planning notes, since the most recent version is in google drive (it should also be noted that the file in google drive has not changed in ~9 months).


rabinowitz:
I have similarly been bothered by the eyesore that is having two planning folders and never knowing which one is correct. However, I think having both as reference for just a little while longer while those who are working on this make larger changes will be helpful. I plan to join that group this semester.

@echang264 could easily override my opinion though since he's been working with this.

rabinowitz (approved):
LGTM! Clarified in person that this is old code written at last year's competition with little relevance to today.

PR #212, opened 2020-09-13
Merge docker repo into code repo

muthuArivoli:
Need to add the docker bot token as a secret (probably an organizational secret) before merging.

#213 provides proof that the CI will take the `build-without-docker` pathway when only files that are not in the docker folder are modified. [This repo](https://github.com/muthuArivoli/robosub-ros-test) provides proof that stuff works with pushes to master (look at the latest two commits for examples of each pathway).

The CI will not work well for PRs not targeting master that modify the docker stuff, since the CI will always test against a merge with master. I will work on this in the future, since that will require some complicated git stuff or figuring out the intricacies of the github API.

Also note the CI will fail if a fork makes a PR modifying the docker stuff, but I think that should be fine since we don't use forks anyway.


neildhar (approved):
amazing

rabinowitz (approved):
Looks amazing! Will get you that secret soon.

PR #222, opened 2020-10-03
Create simulation package and cleanup

puzzler7:
This moves our existing simulation code into a ROS package and cleans up the code inside of it. It also fixes the DVL publishing for the simulation.


muthuArivoli (changes requested):
Here are a couple notes since I can't comment on some files.

What is `comm_code_old.lua`? If it's not going to be used, then please delete it.

The gate model, `Gate.ttm` contains code embedded inside its binary file. I think you can change it to the way you did cthullu, with a single require at the top. Also, try to make capitalization consistent, since cthullu is lowercase, so this should also probably be lowercase.


What is the model `PhysicsRob.ttm`? If it's the same as cthullu, then delete it. Otherwise, remove the code embedded inside it like for `Gate.ttm`.

What is the difference between `RobObstacleCourse.ttt` and `gate_scene.ttt`? Again, if it's not needed then delete it. Also keep capitalization consistent. Also, there is code embedded in the binary file, so move that into its own lua file if you want to keep it.

landside/catkin_ws/src/simulation/scripts/squareCommand.py
```diff
@@ -1,7 +1,8 @@
-import thread
+#!/usr/bin/env python
 import time
 
 import rospy
+import thread
```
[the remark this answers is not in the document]
puzzler7: Done - threading was used in the first place solely for a "press enter to kill", which is very unnecessary. As such, much of the script has remained the same.

puzzler7:
Requested changes have been made.

I replaced the RobWithCam and Gate files entirely, making new ones without embedded code from the currently functional gate_scene. The code that was inside Gate was defunct CV code, and so it has been removed.

As you correctly noticed, many files are old duplicates, and so have been removed.

landside/catkin_ws/src/simulation/scripts/squareCommand.py
```diff
@@ -1,19 +1,13 @@
...
-def input_thread(a_list):
-    input()
-    a_list.append(True)
-
-
 if __name__ == "__main__":
     pub = rospy.Publisher("/sim/move", Float32MultiArray, queue_size=10)
     rospy.init_node("move")
-    start = time.time()
+    start = rospy.get_rostime().secs
```
muthuArivoli, line 10: Just set this to rospy.get_rostime(), subtracting to ros times gives you a duration which you can use to_sec() to convert to a floating point seconds.

landside/catkin_ws/src/simulation/scripts/squareCommand.py
```diff
@@ -23,14 +17,13 @@ def input_thread(a_list):
     ct = 0
     toPub = forwards
     a_list = []
-    thread.start_new_thread(input_thread, (a_list,))
-    print("Started sending. Press enter to kill.")
     while not a_list:
```
muthuArivoli, line 20: Use `rospy.is_shutdown()`

landside/catkin_ws/src/simulation/scripts/squareCommand.py
```diff
@@ -23,14 +17,13 @@ def input_thread(a_list):
...
         # now = rospy.get_rostime()
         # rospy.loginfo("Current time %i %i", now.secs, now.nsecs)
         # print(now)
         data = Float32MultiArray()
-        if time.time() - start > 5:
+
+        if rospy.get_rostime().secs - start > 5:
```
muthuArivoli, line 26: Do `(rospy.get_rostime() - start).to_secs()`

muthuArivoli (changes requested):
All comments are about the square command script. Don't worry about it, I'll work on it tonight and merge it before tomorrow, so people can learn the new structure.

muthuArivoli (approved):
LGTM

PR #224, opened 2020-10-08
Update README.md

nickchak21:
Added a command about sourcing the ROS commands in onboard so you can use roscore in simulation.


muthuArivoli (changes requested):
Not needed

PR #225, opened 2020-10-08
Set line endings for bash scripts

nickchak21:
Set line endings in the gitattributes for bash scripts. Also cleared up where build.sh is located in Docker container.


README.md
```diff
@@ -157,7 +157,7 @@ The following needs to be done once at the beginning, and then later only when m
 #### Executing Build Script
 We use a custom build script that simplifies the process of building and overlaying catkin workspaces.
 
-To build our workspaces for ROS, in the `robosub-ros` directory, execute
+To build our workspaces for ROS, in the `/dev/robosub-ros` directory, execute
```
muthuArivoli, line 160: This is technically incorrect. It should be either `~/dev/robosub-ros` or `/root/dev/robosub-ros`

muthuArivoli (changes requested):
See change

README.md
nickchak21: I made the change

rabinowitz:
I like the folder path clarification.

I don't see the autocrlf change in the Files Changed tab on this PR, but it was made in one of the commits. Is it possible you undid it by mistake?

Further, I think we should be able solve the problem without changing the readme, but instead by adding to the `.gitattributes` file, as you can see I did [here](https://github.com/DukeRobotics/robosub-ros/blob/task-planning/.gitattributes#L1) in the task-planning branch.

muthuArivoli:
We removed autocrlf, since we don't want to change a user's system git settings just for our project. Agreed that the gitattributes is the correct solution, will add that to this PR.

rabinowitz (approved):
LGTM

muthuArivoli (approved):
LGTM, merging now. Future README updates should only be package specific, so bundle them up in an existing or upcoming PR that is targeting that package only.

PR #230, opened 2020-10-17
Simplify interface for servos and optimize Arduino memory

### LINKED ISSUE #227: Use the standard servo interface for offboard comms and optimize memory
## Background
Our current offboard comms code (the Arduino code) uses quite a bit of memory, at least compared to the maximum memory on the Nano. We are currently running into stability problems, so we definitely need to reduce the memory footprint, especially if there are any plans to add code for more sensors.

Also, our current implementation of the MultiplexedESC and MultiplexedServo does not really make sense, since it does not really follow the single-responsibility principle, since they are both doing the same basic thing. Thus, we should probably simplify the code. The inheritance that we use could also be affecting the memory as well, so its probably better to rewrite the code to not use inheritance.

## Proposed Solution
With regards to the standard interface, we will use the API provided by the standard Arduino servo library, located [here](https://www.arduino.cc/reference/en/libraries/servo/). This will allow use to easily change our code to use the standard servo library, if we want to test new code without using the multiplexer. Depending on memory constraints, we might omit some of the less used functions, like read().

To optimize memory, we will reduce the buffer size used by the node handler. We will also change all of our integer variables to type uint8_t, an 8-bit integer, since that will save us a couple of bytes. We will also stop using pointers, since that confuses the memory counter that the Arduino calculates at compile time. Finally, we will look into smaller optimizations, like making sure we are optimizing alignment of the instance variables.

muthuArivoli:
Closes #227

Memory Usage

Master: 
Sketch uses 20098 bytes (65%) of program storage space. Maximum is 30720 bytes.
Global variables use 1729 bytes (84%) of dynamic memory, leaving 319 bytes for local variables. Maximum is 2048 bytes.

PR:
Sketch uses 19702 bytes (61%) of program storage space. Maximum is 32256 bytes.
Global variables use 1708 bytes (83%) of dynamic memory, leaving 340 bytes for local variables. Maximum is 2048 bytes.


onboard/catkin_ws/src/offboard_comms/Arduino Sketchbooks/ThrusterServoArduino/MultiplexedServo.cpp
```diff
@@ -2,15 +2,51 @@
 #include <Wire.h>
 #include "Adafruit_PWMServoDriver.h"
 #include "MultiplexedServo.h"
-#include "MultiplexedObject.h"
 
-MultiplexedServo::MultiplexedServo(Adafruit_PWMServoDriver *_multiplexer,int _num):MultiplexedObject(_multiplexer, _num){}
+MultiplexedServo::MultiplexedServo(Adafruit_PWMServoDriver *_multiplexer):multiplexer(_multiplexer){
```
neildhar, line 6: I would recommend making this just a default constructor (doesn't take any arguments), and then passing in the multiplexer value in a separate `initialise` function. Then you can avoid the heap allocation using `new` in `setup`, by just making `thrusters` and `servos` an array of `MultiplexedServo` instead of an array of `MultiplexedServo*`.

onboard/catkin_ws/src/offboard_comms/Arduino Sketchbooks/ThrusterServoArduino/MultiplexedServo.cpp
```diff
@@ -2,15 +2,51 @@
...
+
+void MultiplexedServo::attach(uint8_t _pin){
+  attach(_pin, 544, 2400);
+}
+
+void MultiplexedServo::attach(uint8_t _pin, uint16_t _out_min, uint16_t _out_max){
+  attach(_pin, _out_min, _out_max, 0, 180);
+}
+
+void MultiplexedServo::attach(uint8_t _pin, uint16_t _out_min, uint16_t _out_max, int16_t _in_min, int16_t _in_max){
```
neildhar, line 22: This is somewhat counterintuitive as an API for a servo class, and makes the program harder to read because both the thrusters and servos are now represented as `MultiplexedServo` objects.

The bigger issue though is that you are now storing these range values inside every object, which makes them much larger than they need to be. This would make sense if there was a need to dynamically set the input and output ranges, however, in this case, we have that information at compile time.

It's also important that it be easy to swap out the `MultiplexedServo` class for the built in `Servo` class because that will make it a lot easier to run simple tests. Keeping almost the same API is a big part of that.

I'd recommend keeping the `MultiplexedBasicEsc` class and just having it wrap around a `MultiplexedServo` object, and the former can store the specific constants needed to do conversions.

muthuArivoli:
When you say wrap around, do you mean inheritance or composition? I just did inheritance since it seems like less code, since I don't need to write implementations for all the functions, just the ones that changed.

For the input min/max and output min/max, do you think I should put them in macros, or just leave them as is in the map function.

onboard/catkin_ws/src/offboard_comms/Arduino Sketchbooks/ThrusterServoArduino/MultiplexedBasicESC.h
```diff
@@ -1,15 +1,14 @@
...
+#include "MultiplexedServo.h"
 
-class MultiplexedBasicESC:public MultiplexedObject{
+class MultiplexedBasicESC: public MultiplexedServo{
 public:
-    MultiplexedBasicESC(Adafruit_PWMServoDriver *, int);
-    void initialise();
-    void run(int);
+  void attach(uint8_t);
+  void write(int8_t);
```
neildhar, line 11: This declaration here actually hides the `write` function in MultiplexedServo, instead of overriding it. It doesn't technically matter here since we don't use the two interchangeably, however, it's not considered good practice.

So for instance, a `MultiplexedBasicESC*` would call the `MultiplexedBasicESC` version of the function, but if you cast it to a `MultiplexedServo*`, it would call the `MultiplexedServo` version, this is typically not how you want the inheritance hierarchy to work.

neildhar (commented):
Awesome, these changes look great, have you measured the new memory usage numbers?

I meant composition, i.e. having a MultiplexedBasicESC store a MultiplexedServo as a member and then call through to that.  Inheritance adds some extra complications here and it's not necessary because there's very little code being duplicated. We're also not using these classes interchangeably.

There's definitely no need for macros here.

muthuArivoli:
Yeah, I editing the top comment with the new memory usage stats, at least as I add new commits. It's saying that we have gained 21 bytes from this new sketch, and I'm hoping that since I declared some of the variables in the parameters in the class methods as 8 bit integers, that will also help with not overflowing.

I've also just received the Nano Every, which has 3x more dmem, and when I compile with that, I get 17% memory usage, so hopefully that works. Some people were claiming there was an issue with running rosserial with it though, in some issue in the rosserial repo, so we'll see where that goes.

neildhar (approved):
Looks great

PR #233, opened 2020-10-30
Clean argument for build script

### LINKED ISSUE #228: Add clean argument to build script
## Background
As evidenced by recent meetings, it takes quite a while to clean the workspaces, since you have to cd into a nested directory, and cd back multiple times. Thus, the clean argument would speed up and simplify that process.

## Proposed Design
The clean argument will clean all three workspaces, no matter which container you are located in. The reason for that is if you have already built a workspace that depends on core, and then clean core, without cleaning that workspace, that workspace will not work anymore. Thus, if we were to only clean onboard and core in the onboard container, with landside already being built, we would probably get a weird error in landside. 

If someone wants to only clean the onboard or landside workspace, then they can simply cd into the appropriate workspace and run catkin clean. This clean argument should only be used when you want to clean the core workspace, since that requires the complicated multiple cd arguments that take time, so this saves time.

## Test Plan
We will test using the clean command to clean the workspace when everything is built. We will also test the clean command when some and all of the workspaces are not built. We will ensure that errors do not propagate to the shell and cause undefined behavior. All tests will be run in both containers.

muthuArivoli:
Closes #228.
Although it's not mentioned in the issue, I decided to add the option of using a second argument to clean a specific workspace, just for convenience.

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,9 +224,13 @@ Here are some common launch configurations for both pool and local testing.
         ```
 
 ### Cleaning Build
-To clean the build outputs from a workspace (build, devel, and logs folders), in `robosub-ros/<workspace>/catkin_ws`, execute
+To clean the build outputs from a workspace (build, devel, and logs folders), in the `~/dev/robosub-ros` directory, execute
 ```bash
-catkin clean
+./build.sh clean <workspace>
+```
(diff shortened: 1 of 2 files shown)


ShaanGondalia (approved):
LGTM

PR #236, opened 2020-11-05
Fix CI for PRs that do not target master

### LINKED ISSUE #234: Fix CI for PRs that do not target master
## Background
Currently, the CI does not work properly for PRs that do not target master, since the base branch is always assumed to be master (which is then merged into the branch from the PR for testing), see [here](https://github.com/DukeRobotics/robosub-ros/blob/7263c7ae947787978870c0420f9bb4d2b838a5ce/docker/clone-and-build.test.yml#L8). This could potentially break, and just isn't representative of what happens when the branch is merged. 

This was partly discussed in #215.

## Implementation
To fix this, we would need to pass in another environment variable to the test step in both the onboard-docker and landside-docker jobs. This environment variable would be the base branch for PRs, or master for pushes. To access the base branch on PRs, one can use `github.event.pull_request.base.sha`, and simply merge that instead of merging master.

muthuArivoli:
Closes #234 
Also forces the docker CI to run whenever the workflows directory changes.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,6 +22,7 @@ jobs:
         filters: |
           docker:
             - 'docker/**'
+            - '.github/workflows/**'
 
   core-docker:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


ShaanGondalia (approved):
LGTM

PR #238, opened 2020-12-18
Updated CV README.md and publishing height and width of images

tylerfeldman321:
Updated the detection script and the CVObject to publish the height and width of the images that are being used for detection. Made the README clearer and re-ordered the sections


ShaanGondalia (approved):
LGTM. Squash commits before merging.

PR #240, opened 2021-01-27
Controls v2

RyanWolf01:
Merging version 2 of controls with velocity control and nested PID loop


onboard/catkin_ws/src/controls/scripts/desired_state.py
```diff
@@ -74,12 +89,12 @@ def run(self):
         while not rospy.is_shutdown():
             rate.sleep()
 
-            if self.pose and self.powers:
+            if (self.pose and self.twist) or (self.pose and self.power) or (self.twist and self.power):
                 # More than one seen in one update cycle, so warn and continue
                 rospy.logerr("===> Controls received both position and power! Halting robot. <===")
```
muthuArivoli, line 94: Do these log messages need to be updated for velocity?

onboard/catkin_ws/src/controls/scripts/desired_state.py
```diff
@@ -99,34 +114,16 @@ def run(self):
...
+            elif self.twist:
+                self.enable_loops()
+                self.disable_pos_loop()
+                utils.publish_data_dictionary(self.pub_vel, utils.get_axes(), self.twist)
+                self.twist = None
+
+            elif self.power:
+                self.disable_loops()
+                utils.publish_data_dictionary(self.pub_power, utils.get_axes(), self.power)
+                self.power = None
```
muthuArivoli, line 126: This code has the side effect of no longer holding state on power control. Is this intended? I think joystick would prefer if you did maintain state like before, cause who wants try to figure out the exact amount by which to hold the joystick down for the robot to not float straight up.

onboard/catkin_ws/src/controls/README.md
```diff
@@ -91,7 +94,7 @@ Thruster speeds are published for the use of the simulation and Arduino.
     + Type: controls.msg/ThrusterSpeeds
```
muthuArivoli, line 94: This should be `custom_msgs/ThrusterSpeeds`

muthuArivoli (changes requested):
Just a first pass through. The main question I have is that it seems as though power control (i.e joystick) will no longer hold state on the axes that are not being sent power to. Is this intended? I think it might be better to hold state on power control so that the robot just doesn't bob up if you let go of the z joystick.

onboard/catkin_ws/src/controls/scripts/test_launch.py
```diff
@@ -0,0 +1,40 @@
+import roslaunch
```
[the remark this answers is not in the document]
nickchak21: It is used in a few places in the file.

onboard/catkin_ws/src/controls/scripts/desired_state.py
```diff
@@ -99,34 +114,16 @@ def run(self):
...
-                if self.powers['z'] != 0:
-                    utils.publish_data_constant(self.pub_pos_enable, ['z'], False)
-
-                utils.publish_data_dictionary(self.pub_power, utils.get_axes(), self.powers)
-                # Publish current state to the desired state for PID
-                utils.publish_data_dictionary(self.pub_pos, utils.get_axes(), self.hold)
-                self.powers = None
+            elif self.twist:
+                self.enable_loops()
+                self.disable_pos_loop()
```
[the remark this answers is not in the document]
ShaanGondalia: I disagree. The solution you propose has same issue of queue size limits, as we would first have to disable both loops with a self.disable_loops() call then only enable the vel loop. This can only be avoided we have a disable/enable function for both pos and vel.

muthuArivoli (approved):
LGTM

PR #241, opened 2021-01-27
Add updated info to the CV README

alankbi:
### List of Changes

* Add info explaining how only the highest-confidence prediction of each label type is published
* Add info explaining how a message with label field 'none' is published if a model is enabled that detects no objects in a frame


nickchak21 (approved):
LGTM

PR #243, opened 2021-01-27
Specify in CV README that the top-left corner is the origin 

alankbi:
### List of Changes

* Specify in the CV README that (0, 0) is the top-left corner of the image when publishing results


rabinowitz (approved):
LGTM

PR #248, opened 2021-02-06
Daily build

### LINKED ISSUE #242: Add automated daily docker pull
Docker recently changed their policy according to [this](
https://www.docker.com/blog/docker-hub-image-retention-policy-delayed-and-subscription-updates/#:~:text=Two%20months%20ago%2C%20we%20announced,after%20six%20months%20of%20inactivity).

To ensure that our images are not deleted, we will need to pull from them at least once every six months. Automating this will ensure our images never get deleted since we went on break or something. Should be simple to do with github actions.

muthuArivoli:
Closes #242.
In addition to pulling the image, also builds our code as well.


rabinowitz (approved):
I'd recommend once a week, but if you don't think we'll run out of our GitHub action credits, then this is fine.

PR #249, opened 2021-02-06
Update pip install around version 21.1

muthuArivoli:
Updates pip install for to use apt instead of 21.1. This fixes the CI that has been broken on docker builds for the last month.
Ties CI down to Ubuntu 18.04.
Ties docker down to 19.03.


rabinowitz (approved):
LGTM!

PR #250, opened 2021-02-10
Acoustic test2

reedchen19:
Added model_new_python.py to Acoustic test2 which is the only file that needs to be added. The Acoustic test2 branch can be deleted since no one is using it anymore.


muthuArivoli:
Will merge manually due to conflicts.

PR #255, opened 2021-03-03
Acoustics

### LINKED ISSUE #229: Acoustics package tracking issue
This issue serves as a tracking issue for the various components that need to be created for acoustics.

## To-do
- [ ] Sampling
    - [ ] Review the saleae CLI and rewrite as needed
    - [ ] Design interface to expose to the main node
    - [ ] Integrate the new saleae code into ROS
    - [ ] Test with actual hardware (saleae for now)
- [ ] Guessing (Related PRs: #205) 
    - [ ] Integrate into ROS
    - [ ] Review, and possibly rewrite, feedback used in action
    - [ ] Use NumPy to speed up computations
- [ ] Processing (Related PRs: #198)
    - [ ] Integrate into ROS
    - [ ] Simplify inputs into package (like the filename and version) and overall design
    - [ ] Review, and possibly rewrite, feedback used in action
    - [ ] Use NumPy to speed up computations
- [ ] Main Wrapper (Related PRs: #226) 
    - [ ] Discuss the best interface to expose with task planning
    - [ ] Create client for the three other components
    - [ ] Create server for task planning

## Background
This package will handle the acoustics task in the competition. It will integrate the sampling and processing together, and provide an easily used interface that task planning can use when they want to call this task.

## Proposed Design
There will be 4 nodes in the acoustics package. The sampling node will be responsible for communicating with our sampling hardware to get the required data. For now, we will assume that we are using the saleae, and if/once we start using an FPGA, we can rewrite this node to use the FPGA. The guessing node will be responsible for taking in some acoustics data and generating the octant that the pinger is located in. The processing node will take in some different acoustics data and the guess generated from the guessing node, and will generate the horizonal angle with respect to the robot that the pinger is located at. The main wrapper node will call the other three nodes, in a sequential fashion, and this is the node that task planning will communicate with to start the process. It will serve as the main server for the acoustics package as a whole, and clients for the other three parts of the package.

The external interface will be an action that contains the data:
```yaml
# Input 
# Target Frequency
int tar_f
---
# Output
# Final horizontal angle
float64 hz_angle 
---
# Feedback
int curr_stage
int total_stage
```
Task planning will be responsible for supplying the target frequency, since that is defined during the competition, depending on the quadrant we are running the robot in. The feedback will be the total number of stages that need to execute, and the current stage that is executing. The stages that we currently expect are:

Before action is called: Robot stabilizes and holds position (we are not responsible for this, task planning is) 

1. Sampling with hydrophone set 2 (guessing hydrophones)
2. Guessing with data from step 1
3. Sampling with hydrophone set 1 (processing hydrophones)
4. Processing with data from steps 2, 3

Again, as noted above, the main wrapper node will call all of these steps. We may investigate pipelining this process to get multiple sets of data to test accuracy, while preserving performance. We might also try to run steps 2, 3 in parallel.

## Test Plan
We will first test the guessing and processing node individually to see if it generates correct outputs, based on data we have already obtained. We will then run some tests on our sampling node using the actual hardware in the lab. We will then finally test all nodes combined in the main wrapper node, and with that we will confirm that all components work.

muthuArivoli:
Closes #229 

Adds the standard ROS interfaces and 5 nodes for acoustics: the wrapper, saleae, data_sim, guessing, and processing, as noted in the issue.

Note data_sim was not mentioned in the issue, it is simply a replacement for sampling (the saleae), so that we can simulate data in software, either for simulation or for testing.

Opening up now, but we still need to do some testing for the whole system, create some launch files, add a readme, add docker deps (and figure out a way to install Logic), and integrate data_sim into the acoustics wrapper.


ShaanGondalia:
Eventually there should be a README for the package and the main README should be updated (not prudent). Otherwise LGTM.

ShaanGondalia (approved):

PR #256, opened 2021-03-06
Make desired pose global, make desired twist local

nickchak21:
All desired poses should be treated as global (need to transform), and all desired twists should be treated as local (no need to transform).


muthuArivoli (approved):
LGTM

PR #269, opened 2021-03-20
Code Formatting Script

### LINKED ISSUE #263: Code formatting script
## Background
No one likes the linter since it's very annoying, but we need it to deal with people using tabs instead of spaces and just bad code formatting in general. Right now people have to manually check the ci, fix any linter errors, during which they inevitably create more linter errors, and the cycle of life goes on.

## Task
I'm 95% sure that this can be done in a single line of code:
`autopep8 -r --in-place --aggressive --aggressive .`
but this should be tested and wrapped in a script, with dependencies added to the dockerfile. This script should be called `format-code.sh`. The script should be located in a folder called scripts in the root of the repo (you may have to create this).

ShaanGondalia:
Closes #263

### DIFF
--- a/docker/core/Dockerfile
+++ b/docker/core/Dockerfile
@@ -79,6 +79,9 @@ RUN ssh-keygen -A
 
 ### Final Setup ###
 
+# Install code formatting packages
+RUN pip install autopep8
+
 # Add catkin ws
 RUN mkdir -p /root/dev/robosub-ros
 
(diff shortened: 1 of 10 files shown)


ShaanGondalia:
After some preliminary testing, it seems like the standards that autopep8 uses are not exactly the same as the standards in the CI. Not sure if there's anything we can do about this.

muthuArivoli (approved):
LGTM

PR #271, opened 2021-03-23
wrapper and data generator

### LINKED ISSUE #264: Unify the acoustics data simulator and saleae interface
## Background
The acoustics code has two ways of creating data - via hardware input with the saleae action server, or using the data_sim action server. Currently, only the saleae action is integrated in the acoustics wrapper (the external API provided for acoustics), but we should have also integrate the data_sim action server in the wrapper, which will allow us to easily call the entire acoustics system (data_sim + guessing + processing) with a simplified API, instead of manually calling each of the action servers and transferring data by hand.

## Task
To do this, the wrapper should be modified to call the proper action server (data_sim or saleae) based on a parameter passed into the node. This parameter would be in the launch file, and should be easily changeable, like the sim arg in rest of the codebase (perhaps this could use that same sim arg). Since the data_sim action needs the additional parameter of the location of the pinger to simulate, the wrapper node server could be modified to also use a different action type based on the sim parameter. An additional action type would be required, similar to the currently existing AcousticsWrapper.action, but this would also include a field for the desired location of the pinger in the goal part of the action.

reedchen19:
Closes #264


core/catkin_ws/src/custom_msgs/action/AcousticsWrapper.action
```diff
@@ -3,6 +3,12 @@
 # Target frequency
 uint32 tar_f
 
+# Use Saleae or Generated Data
+bool gen_data
+
+# Location of the pinger (only if using generated data)
+geometry_msgs/Point location
+
```
muthuArivoli, line 11: Remove these, you should not be modifying the AcousticsWrapper action, since this is the external interface with which task planning communicates with acoustics, so they should not be confused by this. Instead, create a new action, that will be used with the wrapper if using generated data.

onboard/catkin_ws/src/acoustics/scripts/acoustics_wrapper.py
```diff
@@ -24,10 +26,12 @@ def __init__(self):
         self.client_sampling = actionlib.SimpleActionClient('call_saleae', SaleaeAction)
```
muthuArivoli, line 26: For this, you need to make sure that you only start either the saleae client or the generate_data client, not both. Read in a ros parameter that determines which you are using, and only start the one.

onboard/catkin_ws/src/acoustics/scripts/acoustics_wrapper.py
```diff
@@ -24,10 +26,12 @@ def __init__(self):
...
         self.client_guess = actionlib.SimpleActionClient('guess_acoustics', AcousticsGuessAction)
         self.client_processing = actionlib.SimpleActionClient('process_acoustics', AcousticsProcessingAction)
+        self.client_data = actionlib.SimpleActionClient('generate_data', AcousticsDataAction)
 
         self.client_sampling.wait_for_server()
         self.client_guess.wait_for_server()
         self.client_processing.wait_for_server()
+        self.client_data.wait_for_server()
 
         self.server = actionlib.SimpleActionServer(self.ACTION_NAME, AcousticsWrapperAction, self.execute, False)
```
muthuArivoli, line 36: Here, you should read in that ros parameter and create the server to be of type AcousticsWrapperAction or your custom wrapper type for use with data_gen.

onboard/catkin_ws/src/acoustics/scripts/acoustics_wrapper.py
```diff
@@ -56,6 +60,20 @@ def saleae_sampling(self, hydrophone_set, save_name):
         self.client_sampling.wait_for_result()
         return self.client_processing.get_result()
 
+    def generate_data(self, hydrophone_set, wrapper_goal):
+        goal = AcousticsDataGoal()
+        goal.hydrophone_set = hydrophone_set
+        if hydrophone_set == 1:
+            goal.samp_f = self.GUESS_SAMPLE_FREQ
+        else:
+            goal.samp_f = self.PROCESSING_SAMPLE_FREQ
```
muthuArivoli, line 69: Create a dictionary and use it to simplify this code and remove if statements.

onboard/catkin_ws/src/acoustics/scripts/acoustics_wrapper.py
```diff
@@ -56,6 +60,20 @@ def saleae_sampling(self, hydrophone_set, save_name):
...
+        goal.tar_f = wrapper_goal.tar_f
+        goal.location = wrapper_goal.location
+        self.client_data.send_goal(goal)
+        self.client_data.wait_for_result()
+        return self.client_datae.get_result()
```
muthuArivoli, line 74: client_data

onboard/catkin_ws/src/acoustics/scripts/acoustics_wrapper.py
```diff
@@ -76,19 +94,31 @@ def get_processing_angle(self, wrapper_goal, processing_files, guess):
...
     def execute(self, wrapper_goal):
+
         time_now = datetime.now()
         guess_export_name = "date_" + time_now.strftime("%m_%d_%Y_%H_%M_%S") + "_guess"
         processing_export_name = "date_" + time_now.strftime("%m_%d_%Y_%H_%M_%S") + "_processing"
 
-        self.publish_feedback(0, "Starting sampling for guess")
-        guess_sampling = self.saleae_sampling(1, guess_export_name)
-        self.publish_feedback(1, "Sampling for guess complete, starting guess")
+        if wrapper_goal.gen_data == True:
```
muthuArivoli, line 102: When you remove this from the wrapper msg, this will become a param. What you might want to do is set a variable that is equal to the param in the init function, so that you don't have to continue trying to get the param, you can just read the instance variable.

muthuArivoli: Edit: You'll be able to remove this if statement completely if you use pass the callback fn as a lambda into this execute fn. See next comment

onboard/catkin_ws/src/acoustics/scripts/acoustics_wrapper.py
```diff
@@ -76,19 +94,31 @@ def get_processing_angle(self, wrapper_goal, processing_files, guess):
...
 
-        guess = self.get_guess(wrapper_goal, guess_sampling.file_paths)
-        self.publish_feedback(2, "Guess complete, starting sampling for processing")
+            guess = self.get_guess(wrapper_goal, guess_sampling.file_paths)
+            self.publish_feedback(2, "Guess complete, starting sampling for processing")
 
-        processing_sampling = self.saleae_sampling(2, processing_export_name)
-        self.publish_feedback(3, "Sampling for processing complete, starting processing")
+            processing_sampling = self.saleae_sampling(2, processing_export_name)
+            self.publish_feedback(3, "Sampling for processing complete, starting processing")
```
muthuArivoli, line 121: This needs to be redesigned, you have code duplication here in guessing and processing occurring in the if statements. You should pass in the fn you want to call as a argument into this execute function, and when you create the action server, assign it there using a lambda to forward the argument based on what type of server you're creating, which was based on that param you read in.

onboard/catkin_ws/src/acoustics/scripts/data_sim.py
```diff
@@ -8,28 +8,28 @@
...
-        if self.hydrophone == 'cheap':
+        self.hydrophone_set = hydrophone_set
+        if self.hydrophone_set == 1:
             space = .3
             self.hp = [[0, 0, 0],
                        [space, 0, 0],
                        [0, space, 0],
                        [0, 0, space]]
-        elif self.hydrophone == 'expensive':
+        elif self.hydrophone_set == 2:
```
muthuArivoli, line 24: This should use an enum (create two constants, one representing cheap and one expensive) defined in the message and compare to it there. I know I missed this in the saleae action, so change it there as well. Also, when you call the function, in the wrapper, you should use this enum instead of using the number 1 or 2

muthuArivoli (changes requested):
Code needs significant redesign, see comments. I'll can handle the easier stuff like the enum, but you need to do everything else like fixing the wrapper script to use a ros parameter, and not modifying the external interface by creating your own wrapper action.

muthuArivoli (approved):
LGTM, data_sim integration has been tested, there is still an issue with the processing code printing `invalid acoustics data` for 3/4 of the simulated data files, but will figure that out in a future PR.

PR #274, opened 2021-03-31
Add TopicNames

### LINKED ISSUE #268: Create unified location for topic names via ROS message
## Background
Currently, if we want to change a topic name, we have to hunt down its uses in various locations, which can be irritating (see [this](https://github.com/DukeRobotics/robosub-ros/pull/240#discussion_r573442790)). Thus, having a common location for topic names would help in avoiding such situations. While a YAML config file could work, using constant values via a ROS message is would actually be easier to parse, since we do not need to call an extra library to actually parse a file (as in yaml) and also to find the file (resource retriever). 

## Task
Add a message to custom_msgs called TopicNames that contains string vars for each topic name, and set them to the actual topic name using [constants](http://wiki.ros.org/msg#Constants). In the code itself, you should be able to do something like this:
```python
from custom_msgs.msg import TopicNames

joystick_topic_name = TopicNames.kJoystickTopic
```
Use kName to denote constants.

Brokemia:
The TopicNames message contains constants with all of the different topic names used across the code base. This means any change in the name of a topic only has to be done in one place.

Closes #268

### DIFF
--- a/core/catkin_ws/src/custom_msgs/msg/TopicNames.msg
+++ b/core/catkin_ws/src/custom_msgs/msg/TopicNames.msg
@@ -1,0 +1,76 @@
+# This message contains the names of all of the topics we use
+
+std_msgs/Header header
+
+# Sensors
+# If changed, also change sensor_fusion\params\main.yaml and data_pub\README.md
+string sensors_imu_imu=sensors/imu/imu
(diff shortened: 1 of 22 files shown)


muthuArivoli:
No longer think this is a good idea, our system is too complex for this.

PR #276, opened 2021-04-07
Auto launch testing

### LINKED ISSUE #261: Automated launch-based testing
## Background
Currently, we have no way of actually knowing if our code runs, at least not unless someone manually runs it and ensures it works. We also have no way of determining of whether all of the necessary dependencies have been added to the appropriate Dockerfile for each PR. The python code does not require adding dependencies to the CMakeLists.txt file (since we don't compile anything), thus it would be easy for someone to add Python code with unsatisfied dependencies that does not run, which should ideally be impossible (or at least very hard to do). 

## Task
Ideally, there would be a python script that take a single positional argument denoting the workspace (onboard or landside) and would then search through the necessary folders for all launch files, and would then try to execute them using the Python API for roslaunch. If the launch fails, then the python code would throw an error or if using pytest or other python testing framework, it would cause the test to fail. This script/test would be added to the CI, probably as extra steps after the test build for the onboard and landside jobs. The script should be located in a folder called scripts in the root of the repo (you may have to create this).

Ideally, we should also provide nonpositional arguments to allowlist or blocklist certain packages if we want this to be used outside of the ci, and allow for easy testing for certain packages.

A few questions that need to be answered before starting this task:

- How do we handle scripts that don't have launch files (that use rosrun, like test_state_publisher in controls)? Should we require every script to have launch files? Or is it possible to determine executable files via something like executable bits, or should we have packages declare their rosrun scripts in some config file?
- Certain scripts/launch files cannot be run due to hardware requirements (such as the acoustics launch file), since they don't have the necessary hardware attached (acoustics requires the Saleae attached as of  906fdde). Should we require launch files to use a parameter (perhaps the sim parameter that already exists) to ensure that launch files will be able to run when local testing without hardware? Otherwise, how do we guarantee hardware compatibility in environments without the hardware?

samuel-thompsonn:
Closes #261


muthuArivoli (approved):
LGTM, this has also fixed a lot of bugs found. I've disabled testing arm on the ci, since the hardware emulation doesn't appear to work, but other than that, seems to be fine. In the future, might want to add options to the test scripts controlling things like the wait/timeout duration.

PR #277, opened 2021-04-17
Arduino upload

### LINKED ISSUE #247: System utils package
### Background
This package will be responsible for providing information about the system itself and provide utilities to help us during runtime. Currently, there will be two scripts, with the current tasks being:

- [ ] Publish CPU, GPU (if available), RAM, and disk usage stats.
- [ ] Provide an action to allow us to compile and upload code to the Arduino.

### Interfaces

Memory.msg
```
# Common interface for memory items (RAM, disk, GPU memory)
float used
float total
float percentage
```

SystemUsage.msg
```
# Provides details about usage of the system
float cpu_percent
float cpu_speed

float gpu_percent
float gpu_speed
Memory gpu_memory

Memory ram

Memory disk
```

UploadArduino.action
```
# Goal
string filename
string port
# Other Arduino parameters should be automatically detectable, if not, add them as parameters here.
---
# Result
bool success
---
# Feedback
int curr_stage
int total_stages
string message
```

BrandonBae:
Working Arduino Upload Code
Closes #247

### DIFF
--- a/core/catkin_ws/src/custom_msgs/CMakeLists.txt
+++ b/core/catkin_ws/src/custom_msgs/CMakeLists.txt
@@ -32,6 +32,7 @@ add_action_files(
   AcousticsWrapper.action
   Saleae.action
   AcousticsData.action
+  UploadArduino.action
 )
 
 generate_messages(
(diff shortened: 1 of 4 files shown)


muthuArivoli (approved):
LGTM, tested on NUC with a Nano I took from the lab. An option to only compile and not upload should be added to the script, but that can be added when this is added to the CI (where that is needed).

PR #299, opened 2021-06-19
Integrating Stereo Vision and Object Detection

tylerfeldman321:
This includes the code to integrate stereo vision and object detection together. Uses the bounding box produced by the object detection model to filter the depth map from stereo vision, which is then used to find the depth of the detected object, as well as distance, horizontal, and vertical angles.


tylerfeldman321:
Going to close this for now. Need to test this code on the robot with the synchronized cameras. This also depends on the camera(s) we move forward with.

PR #300, opened 2021-06-19
Updated build.sh script

### LINKED ISSUE #289: Remove workspace argument in build script
## Background
As title states, we can remove the workspace argument in the build script, and simply use the environment variable COMPUTER_TYPE (already set in the container) to detect the appropriate workspace. Note that the current behavior of clean should probably be modified/reversed, so that cleaning all workspaces requires an extra argument (like `./build.sh clean all`), with no argument when only cleaning landside or onboard (`./build.sh clean`. This would allow it to fit in with the new behavior of autodetecting the workspace. 

## Task
The final behavior is as follows:
- `./build.sh` should build the appropriate workspaces (either core+onboard or core+landside).
- `./build.sh clean` should only clean the appropriate workspace (either onboard or landside).
- `./build.sh clean all` should clean all three workspaces (core, onboard, and landside).

Spammy0607:
Updated build.sh script to delete arguments and use COMPUTER_TYPE environmental variable
Closes #289


muthuArivoli (approved):
LGTM

PR #302, opened 2021-09-04
Simulation cv

### LINKED ISSUE #130: Dummy Vision Service
We need to simulate CV by detecting when the robot is looking at an object, and if it is, publish the information to a topic. That information is the position and orientation of the object.

Notes: 
Need to publish an arbitrary number of objects? Or each object gets its own topic? Avoid making a custom message, more hassle for everyone.

Position is center of mass? Orientation is relative to something predefined.

puzzler7:
We've mocked CV within simulation, in a manner that should be extensible for other objects.
Closes #130


landside/catkin_ws/src/simulation/scripts/bounding_box.py
```diff
@@ -0,0 +1,146 @@
...
+        # for i in range(3):
+        # 	pos[i] += cam_pos[i]
+
+        xs = []
+        ys = []
+        for i in range(len(points)):
+            grid_point = self.get_grid_point(points[i])
+            if grid_point != [-1, -1]:
+                xs.append(grid_point[0])
+                ys.append(grid_point[1])
```
muthuArivoli, line 79: I feel like this can be done with a list comprehension, will figure it out

landside/catkin_ws/src/simulation/scripts/bounding_box.py
```diff
@@ -0,0 +1,146 @@
...
+        rel_point[2] = newpoint.z
+        rospy.loginfo("Before: " + str(point) + " After: " + str(newpoint))
+        return rel_point
+
+    def get_box(self, xs, ys):
+        print(xs, ys)
+        if len(xs) == 0:
+            xs = [-1]
+        if len(ys) == 0:
+            ys = [-1]
```
muthuArivoli, line 136: create a new variable and assign it using an if conditional, since we don't want to accidentally modify the old variable (also, should be clearer)

muthuArivoli (changes requested):
First pass, will update with changes myself. Main change is to create a custom message type that will make it cleaner to read.

muthuArivoli (approved):
LGTM

PR #307, opened 2021-11-06
Support for Nano Every

hkder:


onboard/catkin_ws/src/offboard_comms/Arduino Sketchbooks/ThrusterServoArduino/ThrusterServoArduino.ino
```diff
@@ -86,7 +100,7 @@ void loop(){
         servos[i].write(servo_angles[i]);
     }
     pressure_sensor.read();
-    pressure_msg.fluid_pressure = pressure_sensor.pressure(100.0f);
+    pressure_msg.fluid_pressure = pressure_sensor.pressure();
```
muthuArivoli, line 103: I believe this is incorrect, the pressure should be in pascals, but the pressure sensor returns data in mbar, so this 100.0f converts to pascal.

onboard/catkin_ws/src/offboard_comms/scripts/port_finder.sh
```diff
@@ -1,9 +1,9 @@
 #!/bin/sh
 
 sudo dmesg | \
-grep tty | \
-sed -n -e 's/^.* ch341-uart converter now attached to \([^ ]*\).*/\1/p' | \
-tail -n 1 | \
+grep "USB ACM device" | \
+awk '{print $5;}' | \
+sed 's/.$//' | \
```
muthuArivoli, line 6: awk is not the best here, since if there is an extra space or something then it will mess up, sed should be better since its just pattern matching, so try using
```suggestion
grep tty | \
sed -n -e 's/^.* \([^ ]*\): USB ACM device/\1/p' | \
tail -n 1 | \
```

muthuArivoli (changes requested):
Mostly just deleting stuff

muthuArivoli (approved):
Make sure to test if it works one last time before merging, but LGTM!

PR #310, opened 2021-12-03
Added NMS Filtering for Overlapping Multi-Class Predictions

ricohasgithub:
- In `detect.py`, we get raw predictions and pass them into the NMS function in utils.py
 - Created `utils.py`, which contains the function `nms` implementing the NMS filtering algorithm to enable prediction filtering based on bounding box overlaps; for each label class, we find the highest confidence predictions that don't overlap with each other and are above the given confidence threshold
    - NMS implementation taken from: https://github.com/rbgirshick/fast-rcnn/blob/master/lib/utils/nms.py
    - The IoU threshold and raw confidence threshold control which bounding boxes are kept
 - Updated ReadMe with this information and added an example using the cv code to simulate the raw camera feed and run a model on this feed


muthuArivoli (changes requested):
Mostly just nits, would like to see less switching between numpy and pytorch if possible.

Sorry for the delay in review!

muthuArivoli (approved):

PR #311, opened 2022-01-06
GUI

### LINKED ISSUE #292: Add ROS service to launch nodes in the onboard container
## Background
The planned GUI will need to launch nodes/launch files from the landside container on the onboard container. Unfortunately, there does not seem to be a good way of doing this (our current approach uses paramiko to establish SSH, and then runs the command). This service will allow us to take advantage of ROS's distributed features, and allow easy calling of it across the system, using the standard rospy library.

## Task
This should probably live in the execute package. It should have two services and one publisher with the following signatures:

StartLaunch.srv
```
string package
string file
---
string uuid
``` 

StopLaunch.srv
```
string uuid
---
string package
string file
bool success
```

LaunchFeedback.msg - type std_msgs/String.msg

The launch feedback subscription should continuously send data from the launched process (all the logged information). Split it up by newlines, or by some string parsing that ensures the log data makes sense.

### LINKED ISSUE #301: Finalize GUI for robot testing/control
## Background
Using a GUI to interface with the robot instead of a series of terminal windows has obvious time and ease of use advantages. The bulk of the work is done, but a few components need to be finalized before implementation. 

## Task
A few parts have further work required:
- [ ] Fix the sizing of buttons on the UI file to match the possible text displayed
- [ ] Create a checkbox for simulation mode in the UI file and add to module file, which will change commands run for use in simulation testing
- [ ] Get working publishing of twists/poses through publisher in module file instead of current nonfunctional SSH method
- [ ] Make method to start motion.launch file
We have attempted to use paramiko to create an SSH terminal to start a launch file by creating a paramiko channel for the launch file to be run in the background, but this was not functional. Will either need to find the correct way to run it over SSH or complete #292 .

muthuArivoli:
Closes #301
Closes #292

First pass at the GUI. Creates the systems plugin, controls plugin, sensors plugin, and offboard plugin. Here is what it looks like: 
Screenshot (26)

The systems plugin allows users to run, launch, monitor, and delete nodes in the onboard container via a simple UI dialog. It also provides system info like CPU and RAM values.

The controls plugin allows for users to easily enable/disable controls, displays the current desired control value, allows users to set a desired control value, shows the PID constants, and allows users to update and set PID constants on the fly.

The sensors plugin allows users to view the status of various hardware, allows users to control the robot with just a keyboard, displays the current state, and allows users to set/reset the state.

The offboard plugin allows users to view thruster values sent by controls, provides a dialog that allows users to test thrusters easily, provides button to actuate servos, and button to allows users to upload code to the arduino.


nickchak21 (commented):
The controls widget works pretty well. I ran into one main problem: I'm having trouble changing the PID constants. In particular, the "P" values for x,y,z position will not change, no matter what I enter in the text boxes. The "I" and "D" values seem fine though. The "Enable Controls" and "Set Desired Pose/Twist" buttons work correctly, along with their respective text boxes.

landside/catkin_ws/src/gui/src/gui/controls_widget.py
```diff
@@ -0,0 +1,267 @@
...
+        self.desired_power_sub = rospy.Subscriber('controls/desired_power', Twist, self.power_received)
+
+        self.gui_publishing = False
+        self.state_timer = QTimer(self)
+        self.state_timer.timeout.connect(self.check_times)
+        self.state_timer.setInterval(100)
+        self.state_timer.start()
+        self.publishing_timer = QTimer(self)
+        self.publishing_timer.timeout.connect(self.publish_controls)
+        self.publishing_timer.setInterval(100)
```
ShaanGondalia, line 67: This rate is currently too slow for controls. I'm assuming this means the desired state is being published at 10Hz. This is the theoretical refresh rate for controls but timing mismatches cause controls to disable and reenable after every publishing event. Consider bumping this to around 15 Hz

landside/catkin_ws/src/gui/src/gui/controls_widget.py
```diff
@@ -0,0 +1,267 @@
...
+        self.enable_controls_timer.timeout.connect(self.check_enable_controls)
+        self.enable_controls_timer.start(100)
+
+        self.state_times = {
+            'pose': rospy.Time.now(),
+            'twist': rospy.Time.now(),
+            'power': rospy.Time.now()
+        }
+
+        self.desired_pose_sub = rospy.Subscriber('controls/desired_pose', Pose, self.pose_received)
```
ShaanGondalia, line 56: Maybe define all topic names and service names as constants so that they can be changed easily in the future if the namespaces change? Up to you.

landside/catkin_ws/src/gui/src/gui/system_widget.py
```diff
@@ -0,0 +1,144 @@
...
+        self.launch_dialog.reset()
+        self.launch_dialog.open()
+
+    def launch_file(self, launchfile):
+        args = ['sim:=true'] if self.simulation else ['sim:=false']
+        start_launch = rospy.ServiceProxy('start_node', StartLaunch)
+        resp = start_launch('execute', launchfile, args, True)
+        return self.append_to_table(resp.pid, 'execute', launchfile, args[0])
+
+    def delete_launch(self, row_value):
```
ShaanGondalia, line 120: Stopping any running launch file crashes the entire GUI.

ShaanGondalia (changes requested):

landside/catkin_ws/src/gui/src/gui/system_widget.py
ShaanGondalia: This still happens intermittently when running certain launch files. I haven't been able to find the root cause, but it doesn't seem to be an issue with the `remote_launch.py` script. The process shuts down correctly and delete_launch is also successful, but somewhere later down the line the gui segfaults. I really have no clue what it is. To reproduce:

1. Run `roslaunch system_utils remote_launch.launch`
2. Run `roslaunch gui gui.launch`
3. Run simulation and enable simulation mode in gui
4. Start `motion.launch` from gui
5. Publish a desired pose from gui
6. Enable & disable controls
7. Attempting to stop `motion.launch` crashes the gui

Something similar happens to tasks.launch. The only lead I have is that if there is an [error] message in the `roslaunch system_utils remote_launch.launch` tab during shutdown the gui is more likely to crash, but this isn't always the case.

muthuArivoli: Yeah, I wasn't able to figure out what was going on last year, I recommend checking out the logs located under ~/.ros/logs or something like that

ShaanGondalia: Because the gui is functional I'm opting to resolve this conversation . I'll add a backlog item for fixing this bug. The CV team is working on widgets to show camera feeds + bounding boxes as well as a cv control panel.

ShaanGondalia (approved):

PR #312, opened 2022-01-29
CV separate class topics

ricohasgithub:
- In `detection.py` created a dictionary system to organize publisher topics by class name; publishers are listed under `model['publishers']` as `publisher_dict`
 - In `detection.py` updated publish_predictions so that it only publishes to the topic of a given prediction's label (class name)
 - In `detection.py` removed publishing for when no classes are detected
 - In `models.yaml` removed model name from the topic field
 - Updated ReadMe with this information


ShaanGondalia (commented):
LGTM

ShaanGondalia (approved):

PR #317, opened 2022-02-19
Simulation: Add more 3d models for competition props

samuel-thompsonn:
## Features
- Adds the following models for use in CoppeliaSim scenes as .ttm files:
  - GmanBuoy and BootleggerBuoy: Vertical sheet-like rectangular prisms
  - RightPathMarker and LeftPathMarker: Exactly as described in slide 4 of [this competition spec](https://robonation.org/app/uploads/sites/4/2021/10/RoboSub-2022-Task-Ideas.pdf)
  - Octagon
  - Bin: A rectangular prism with guesses of the dimensions
  - Torpedo: A rectangular prism with guesses of the dimensions
  - CYSImage: For the gate task

Also included are relevant STLs for path markers and the octagon.
The README in the models/ folder documents the sizes of all objects that are added.

## Notes
- Some sizes aren't known--like the size of the octagon.
- This is the original job of the add-sim-models branch, but that one got bloated up with also adding interface for sim objects and publishing bounding boxes in a more well-designed way. This branch is exclusively for adding 3D models for constructing task scenes.
- Many of these are just rectangular prisms approximating the actual shape of the object. That should be fine for things that the robot doesn't have to go through since ultimately the robot just sees bounding boxes, but we can always go in and make them higher resolution.


ShaanGondalia (approved):
LGTM

PR #318, opened 2022-02-25
Change fake CV to prefix topic names

Brokemia:
CV publishes objects to topics structured in the following way:
`cv/camera name/object/left`
Simulation was publishing fake CV objects to topics structured in the following way:
`object/left`
This change has simulation publish objects to topics structured in the following way:
`cv/simulation/object/left`


muthuArivoli (approved):
LGTM

samuel-thompsonn (commented):
The other option would be to have the simulation literally publish on the exact same topics, so that task planning can't tell the difference between when the simulation is in use and when we're using real life. What are the advantages and disadvantages of doing things this way versus the way I just proposed?

ShaanGondalia:
Is this change relevant now that we're switching to dedicated stereo cameras?

ShaanGondalia:
We can merge this now, just note that the CV topic name format is likely going to change sometime next semester now that we're using dedicated stereo cameras.

PR #320, opened 2022-08-11
Refactor Controls

### LINKED ISSUE #319: Refactor Controls
## Background
The controls package has incurred some technical debt from the upgrade to nested PID control. Many scripts were written by members over 2 years ago and haven't been touched since. Once these members leave the team, most of the knowledge of these systems won't be transferred. Refactoring and documenting the package will improve clarity and serve as a better base for new members to contribute to.

## Tasks

- [x] Add [Google Python Docstrings](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) to most scripts to explain their general purpose and control flow
- [x] Add TODOs to mark issues/areas of improvement that should be explored further to improve our algorithm (Github Issues addressing these will be created after the refactoring effort)
- [x] Unify PID Launch file format 
- [x] Move ROS PID loop interfacing logic from `desired_state` into a new file called `pid_manager`
    * `pid_manager` will hold the publishers and methods enabling/disabling/publishing to the ROS PID loops
    * `desired_state` will hold logic for handling desired state input, calling `pid_manager` accordingly
- [ ] Deprecate `control_utils`
    * Most of these utility methods are only called once in the entire package. These methods should be moved to the scripts that call them
    * Many util methods will be redundant once `pid_manager` is implemented.
    * Util files like this should be avoided because they split related code into several files, which makes debugging more difficult
    * Can discuss this more if there are arguments for keeping `control_utils`

ShaanGondalia:
Closes #319

Adds documentation for most scripts and refactors code to improve clarity. Adds many TODOs to mark areas of improvement for our controls algorithm that should be looked into in the next semester. We can either to create a catch-all issue for working on controls or separate the efforts into more digestible issues for new members.


ShaanGondalia:
Copying from #319: Decided to keep some utility methods in controls utils, but removed the ones that were only used in a single place. Tested changes in simulation and movement is pretty bad. However, I tested the existing code in simulation as well and got equally bad results, so I think the issue is with simulation, not the refactoring.

Will-Denton (approved):

PR #339, opened 2022-09-07
Migrate CI to docker buildx 

### LINKED ISSUE #291: Migrate to docker buildx v0.5 and docker 20.10 in CI
## Background
We haven't updated buildx and docker in a long time, should do that soon. Also, we should move to using buildx to build the images over using the qemu shortcut, now that it seems like buildx can load images from a local cache.

## Task
Upgrade to the versions listed in the title. Greatly simplify `.github/workflows/build.sh` by using buildx to build (which should come bundled with docker 20.10, the default installed version on the github action runner). Make sure to save and load the image from/to the buildx cache (refer to the docker API on how to do that).

ShaanGondalia:
Closes #291 

Update CI to run on latest ubuntu image (22.04) and migrate build workflows to docker buildx. Buildx migration by itself slightly improved core and onboard build speeds (~1 min). Also uses [cache-to and cache-from](https://docs.docker.com/engine/reference/commandline/buildx_build/#cache-from) to save images to GitHub's cache so that we can skip the saving/loading steps. This cuts down build times a little more, but adds a dependency on the GitHub cache. In the bigger picture, this only saves ~10% of build time, so we can revert back if it isn't worth it.

### DIFF
--- a/.github/workflows/delete-artifacts.yml
+++ b/.github/workflows/delete-artifacts.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   cleanup-docker:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
       - name: Cleanup old artifacts
--- a/.github/workflows/build-artifacts.yml
+++ b/.github/workflows/build-artifacts.yml
@@ -50,6 +50,14 @@ jobs:
       - name: Setup environment and build
         run: |
           ./.github/workflows/build.sh
+      - name: Save image
+        run: |
+          docker save ${IMAGE_NAME} | gzip > ${FILE_NAME}
+      - name: Create artifact from image
+        uses: actions/upload-artifact@v2
+        with:
+          name: core
+          path: ${{ env.FILE_NAME }}
 
   onboard-docker:
     runs-on: ubuntu-22.04
@@ -70,6 +78,15 @@ jobs:
       - uses: actions/checkout@v2
         with:
           ref: ${{ needs.check-comment.outputs.sha }}
+      - name: Get core image
+        uses: actions/download-artifact@v2
+        with:
+          name: core
+          path: core_im
+      - name: Load core image
+        run: |
+          docker load < ./core_im/${CORE_FILE_NAME}
+          rm -rf core_im
       - name: Setup environment and build docker image
         run: ./.github/workflows/build.sh
       - name: Save onboard image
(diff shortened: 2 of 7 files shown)


ShaanGondalia:
Caching actually wasn't working as expected, reverting that change. I don't think local caching is an option because all images build in different jobs.

.github/workflows/delete-artifacts.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 jobs:
   cleanup-docker:
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-latest
```
Will-Denton, line 10: shouldn't this also be 22.04 so it doesnt get out of sync?

ShaanGondalia: No, this can be latest because the ubuntu version doesn't really matter for this workflow. This just runs an action that doesn't have any relation to our robot. The reason I set the other ones to a specific version is because they are responsible for building our images, and we don't want an updated version to break our image CI.

muthuArivoli:
Ok, looking good so far, but the challenge with this issue is that buildx doesn't use the local image cache when building an image, instead it pulls the image in the FROM tag from dockerhub by default. What this means for us is that when building the onboard and landside image, the core image that we just built and saved and then loaded from an artifact isn't actually being used, it's the core image from dockerhub that's being used. 

If you look at the logs for the recent build on this branch [here](https://github.com/DukeRobotics/robosub-ros/runs/8236839941?check_suite_focus=true#step:5:139) and [here](https://github.com/DukeRobotics/robosub-ros/runs/8236839941?check_suite_focus=true#step:5:178) you can see that the core image is being downloaded from dockerhub. If you look at a run from a few days ago on one of your previous PRs [here](https://github.com/DukeRobotics/robosub-ros/runs/8198622296?check_suite_focus=true#step:5:259), you can see the core image is not downloaded, since it's already there (and you can how the jetson-pytorch image is downloaded since it doesn't exist locally in the few lines above it).

Now I know I used the word cache here, note that I'm not sure that's the proper terminology with the docker people (from what I remember the --cache-from and --cache-to arguments did something that I didn't really understand). I think what you want is the --build-context argument found [here](https://docs.docker.com/engine/reference/commandline/buildx_build/#build-context) but I have no idea if or how that works.

Lmk if any of this doesn't make sense.

ShaanGondalia:
> Ok, looking good so far, but the challenge with this issue is that buildx doesn't use the local image cache when building an image, instead it pulls the image in the FROM tag from dockerhub by default. What this means for us is that when building the onboard and landside image, the core image that we just built and saved and then loaded from an artifact isn't actually being used, it's the core image from dockerhub that's being used.

That makes sense. I assumed that the build used the saved core image because it was loaded and listed in `docker images`.

muthuArivoli (commented):
Yeah, so long as it works, docker builder sounds good

muthuArivoli (approved):

PR #343, opened 2022-09-17
CV Cleanup

### LINKED ISSUE #341: CV Package Cleanup
## Background
The scripts in the cv package are sparsely commented and could use docstrings to make it clear what each function does. The `stream_images.py` is also incomplete, so it should be removed.

## Tasks
- [x] Add docstring comments for functions in `detection.py` and `test_images.py`.
- [x] Remove `stream_images.py` because it is incomplete.

tylerfeldman321:
## Description
Closes #341 

- Adds docstring comments that more thoroughly explain cv functions.
- Removes `stream_images.py` since it is incomplete.
- Adds try catch statements in main functions


ShaanGondalia (approved):
LGTM

PR #355, opened 2022-10-06
Cleanup competition code from test_state_publisher.py

### LINKED ISSUE #333: Migrate tasks from controls package
## Background

Due to issues with [smach](http://wiki.ros.org/smach) during competition, most of the competition tasks were written in [`test_state_publisher.py`](https://github.com/DukeRobotics/robosub-ros/blob/master/onboard/catkin_ws/src/controls/scripts/test_state_publisher.py), which is a controls script used to test basic movement. All of these tasks should be migrated out of this file into the task planning package.

## Tasks
- [x] Add competition tasks to task planning files
- [x] Remove competition tasks from `test_state_publisher.py`

Brokemia:
Mostly just copy and paste. At the moment it's just contained in a file in task planning in case we need it later.

Closes #333


ShaanGondalia:
Since the code's already in our git history, can we just delete it from the Repo? I would prefer that over having a dead script in our code base

PR #365, opened 2022-10-22
Added max power/twist check to desired_state.py

### LINKED ISSUE #322: Create desired state safety net
## Background

There is no mechanism for filtering desired state messages to make sure the values are realistic. This can result in some setpoints that lead to unreliable (and dangerous) robot behavior. We should implement safety checks for `desired_power` and `desired_twist` to validate that they are in the range of acceptable values. This will require determining what the acceptable values are (see #321) and updating the documentation accordingly.

## Tasks

- [ ] Add logic in `desired_state` that checks for realistic setpoints, rejecting messages and halting controls if not

Swahaha:
Checked the desired state twist/power by axes to an arbitrary maximum twist/power value that is to be set later.

closes #322

### DIFF
--- a/onboard/catkin_ws/src/controls/scripts/desired_state.py
+++ b/onboard/catkin_ws/src/controls/scripts/desired_state.py
@@ -29,6 +29,10 @@ class DesiredStateHandler:
 
     REFRESH_HZ = 10  # for main loop
 
+    # Max power & Max twist for the controls need values
+    maxpower = None
+    maxtwist = None
+
     # All variables are dictionaries that map directions to their corresponding value (local reference frame)
     pose = None  # Desired pose (position)
     twist = None  # Desired twist (velocity)
@@ -62,7 +66,10 @@ class DesiredStateHandler:
         Args:
             power: ROS Twist message corresponding to desired twists in local reference frame
         """
-        self.twist = utils.parse_twist(twist)
+        parsed_twist = utils.parse_twist(twist)
+        if self.twist_state_safety(parsed_twist):
+            self.twist = parsed_twist
+       
 
     def _on_power_received(self, power):
         """Handler for receiving desired powers. A desired power in a given axis represents the control
@@ -72,7 +79,30 @@ class DesiredStateHandler:
         Args:
             power: ROS Twist message corresponding to desired powers in local reference frame
         """
-        self.power = utils.parse_twist(power)
+        parsed_power = utils.parse_twist(power)
+        if self.power_state_safety(parsed_power):
+            self.power = parsed_power
+
+    def power_state_safety(self, power):
+        # Compares power with controller limits
+        return_status = True
+        for axes in utils.get_axes():
+            if power[axes] > self.maxpower[axes]:
+                rospy.logerr("===> Desired power exceeds maximum power in ", axes, "! Halting robot. <===")
+                self.pid_manager.soft_estop()
+                return_status = False
+        return return_status
+            
+    def twist_state_safety(self, twist):
+        # Compares twist with controller limits
+        return_status = True
+        for axes in utils.get_axes():
+            if twist[axes] > self.maxtwist[axes]:
+                rospy.logerr("===> Desired twist exceeds maximum twist in ", axes, "! Halting robot <===")
+                self.pid_manager.soft_estop()
+                return_status = False
+        return return_status
+
 
     def _reset_data(self):
         """Resets all desired state data"""
--- a/onboard/catkin_ws/src/controls/scripts/desired_state.py
+++ b/onboard/catkin_ws/src/controls/scripts/desired_state.py
@@ -30,8 +30,8 @@ class DesiredStateHandler:
     REFRESH_HZ = 10  # for main loop
 
     # Max power & Max twist for the controls need values
-    maxpower = None
-    maxtwist = None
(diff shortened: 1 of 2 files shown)


onboard/catkin_ws/src/controls/scripts/desired_state.py
```diff
@@ -29,6 +29,10 @@ class DesiredStateHandler:
 
     REFRESH_HZ = 10  # for main loop
 
+    # Max power & Max twist for the controls need values
```
ShaanGondalia, line 32: For now, these values can be 1 for x, y, z and 0.5 for roll, pitch, and yaw.

onboard/catkin_ws/src/controls/scripts/desired_state.py
```diff
@@ -72,7 +81,31 @@ def _on_power_received(self, power):
...
         """
-        self.power = utils.parse_twist(power)
+        parsed_power = utils.parse_twist(power)
+        if self.power_state_safety(parsed_power):
+            self.power = parsed_power
+
+    def power_state_safety(self, power):
+        # Compares power with controller limits
+        return_status = True
+        for axes in utils.get_axes():
```
ShaanGondalia, line 91: nit: rename `axes` to either  `axis` or `ax`

onboard/catkin_ws/src/controls/scripts/desired_state.py
```diff
@@ -72,7 +81,31 @@ def _on_power_received(self, power):
...
+            if power[axes] > self.maxpower[axes]:
+                rospy.logerr(
+                    "===> Desired power exceeds maximum power in ", axes, "! Halting robot. <===")
+                self.pid_manager.soft_estop()
```
ShaanGondalia, line 95: Technically we need to set self.valid=False here. We can refactor self.valid into pid manager to avoid the need to do set self.valid, which I think is a better design. I'll make a commit in the fork that resolves this, so don't worry about it.

onboard/catkin_ws/src/controls/scripts/desired_state.py
```diff
@@ -72,7 +81,31 @@ def _on_power_received(self, power):
...
+                return_status = False
+        return return_status
+
+    def twist_state_safety(self, twist):
```
ShaanGondalia, line 99: Same comments as power_state_safety

onboard/catkin_ws/src/controls/scripts/desired_state.py
```diff
@@ -62,7 +69,9 @@ def _on_twist_received(self, twist):
         Args:
             power: ROS Twist message corresponding to desired twists in local reference frame
         """
-        self.twist = utils.parse_twist(twist)
+        parsed_twist = utils.parse_twist(twist)
+        if self.twist_state_safety(parsed_twist):
```
ShaanGondalia, line 73: I'd like this check to be done in the run loop, not the callbacks. There can be some funky behavior if we do the check in the callback. For instance, if the user is publishing a valid pose and invalid twist at the same time, the twist messages will be rejected, and the pose messages will be accepted, which will result in valid pid position control with soft estops happening at the publishing rate.

To resolve this, I think we can add the checks to lines 153 and 155:
```
elif self.twist & self.twist_state_safety(self.twist):
    # Do velocity control
elif self.power & self.twist_state_safety(self.power):
    # Do power control
```

ShaanGondalia (changes requested):

onboard/catkin_ws/src/controls/scripts/desired_state.py
```diff
@@ -72,7 +81,31 @@ def _on_power_received(self, power):
...
-        self.power = utils.parse_twist(power)
+        parsed_power = utils.parse_twist(power)
+        if self.power_state_safety(parsed_power):
+            self.power = parsed_power
+
+    def power_state_safety(self, power):
+        # Compares power with controller limits
+        return_status = True
+        for axes in utils.get_axes():
+            if power[axes] > self.maxpower[axes]:
```
AndrewCouncil, line 91: Should probably get the absolute value of the number, as negative speeds would not get caught here.

`if abs(power[axes]) > self.maxpower[axes]:`

AndrewCouncil (changes requested):

onboard/catkin_ws/src/controls/scripts/desired_state.py
```diff
@@ -74,6 +80,28 @@ def _on_power_received(self, power):
...
+                return_status = False
+        return return_status
+
+    def _twist_state_safety(self, twist):
+        # Compares twist with controller limits
+        return_status = True
+        for axis in utils.get_axes():
+            if abs(twist[axis]) > self.MAX_TWIST[axis]:
+                rospy.logerr(
+                    "===> Desired twist exceeds maximum twist in ", axis, "! Halting robot <===")
```
ShaanGondalia, line 100: Does this syntax work for logging? I think we would have to use a format string instead:
```suggestion
                    "===> Desired twist exceeds maximum twist in %s! Halting robot <===", axis)
``` 
or 
```suggestion
                    f"===> Desired twist exceeds maximum twist in {axis}! Halting robot <===")
``` 
Also can we change the first sentence of the error message to "Desired twist exceeds maximum <axis> value!"

ShaanGondalia (changes requested):

ShaanGondalia (approved):
LGTM

PR #366, opened 2022-10-22
Transform listener usage improvements

Brokemia:
This changes task planning to use a single TransformListener created in `task_runner.py`, rather than creating a TransformListener in each state machine creation method. We also now properly wait for transforms to become available (for up to 15 seconds before timing out), rather than arbitrarily sleeping for 2 seconds.


ShaanGondalia (approved):
Our discussion of restructuring task planning may cause some future changes, but until then these changes are fine.  Can we create a backlog issue and a plan for documenting and addressing the issues with our current system?

Brokemia:
Done, the issue is #371

PR #373, opened 2022-11-05
Migrate to ROS2

### LINKED ISSUE #288: Automate startup of roscore with systemd
## Background
Ideally, we would like the landside container to be the entrypoint of any human interaction with the stack. The GUI is planned to handle things like launching launch files and nodes and to call services and the like by creating an ssh connection and doing so, but the issue still remains that to launch the GUI, you need roscore already running (so the GUI can't launch roscore itself, since it needed it to run before running). Thus, we should automate this startup at boot of the container, using systemd to make it easy to shutdown/restart when necessary.

## Task
Use this (https://blog.roverrobotics.com/how-to-run-ros-on-startup-bootup/) as a very general guide. Note that it should be heavily modified (only need the roscore.service file, and it should not be forking). Add it to the docker image.

### LINKED ISSUE #350: Migrate to ROS 2
## Background
ROS is no longer receiving updates and noetic will reach EOL by 2025. We need to migrate our stack to ROS 2 eventually to make sure we have access to the latest tools and versions of our dependencies. We should follow the [migration guides](https://docs.ros.org/en/foxy/The-ROS2-Project/Contributing/Migration-Guide-Python.html) as a baseline for updating our code-base. 

There are a few ROS 2 features that complicate the migration process:

- Using [`ament`](https://design.ros2.org/articles/ament.html) and [`colcon`](https://docs.ros.org/en/foxy/Tutorials/Beginner-Client-Libraries/Colcon-Tutorial.html) for builds instead of `catkin`. For our purposes, this involves replacing all `CMakeLists.txt` with a [`setup.py`](https://docs.ros.org/en/foxy/How-To-Guides/Developing-a-ROS-2-Package.html#python-packages) script.
- ROS 2 does not use `roscore` to create a master node, instead it uses [DDS](https://docs.ros.org/en/foxy/Installation/DDS-Implementations.html) to set up peer-to-peer communication. We need to scope out how this will work across multiple machines (onboard and landside). [This tutorial](https://roboticsbackend.com/ros2-multiple-machines-including-raspberry-pi/) is probably a good starting point.
- The ROS API is now `rclpy` instead of `rospy`. The framework for creating nodes changed a bit, so we'll have to modify every script in our code base.

Before we start making efforts to complete the migration, we should conduct a thorough analysis of our ROS dependencies to determine their equivalent packages in ROS 2. Once we've confirmed support or have viable alternatives for all packages, we can start efforts to migrate the rest of our code base. We'll surely run into more hiccups along the way, but for now the following tasks should be a good roadmap for the migration.

## Testing
See #387 

## Important Notes
The general process for migrating a package to ROS2 is:
   - Switch build system to `ament`, replacing `CMakeLists.txt` with `setup.py` and modifying file structure.
   - Switch client library from `rospy` to `rclpy`
   - Update pub/sub syntax
   - Rewrite launch files in python
   - Update README.md

## Tasks
- [x] Create current ROS dependency list with ROS 2 equivalents
- [x] Build proof-of-concept to verify cross-machine functionality
   - [ ] Update `setup_network.sh` to remove ROS master and replace with domain id
   - [x] Test onboard/land-side connection on local machine
   - [ ] Test onboard/land-side connection from robot to local machine
- [ ] Migrate all "easy" packages (landside first, then onboard)
   - [x] landside
       - [x] joystick
       - [x] camera_view
       - [x] simulation
   - [ ] onboard
       - [x] acoustics
       - [x] avt_camera
       - [x] cv
       - [x] data_pub
       - [ ] execute
       - [x] robosub_description
       - [x] sensor_fusion
       - [x] static_transforms
       - [x] system_utils
- [ ] Migrate all "hard" packages (controls, offboard_comms, task planning, gui?)
   - [x] Dereference `pid` package in controls, rewriting PID logic
   - [ ] Rewrite current state of task planning to use executive-smach or some other tool
   - [ ] Implement `micro-ros` in place of rosserial, which involves swapping hardware (See #375)
   - [ ] Once everything is updated, we can migrate `gui`.
- [x] Update `build.sh` to use `colcon` instead of `catkin`
- [ ] Update docker images with ROS 2 dependencies (base image and packages)
   - [ ] Update pytorch installation version and `jetson-pytorch` image? Not sure if cv is using pytorch anymore.
- [x] Update Github CI (particularly launch file testing and building)
- [ ] Test each package on Cthulhu. See #387 
- [ ] Make a ROS2 knowledge transfer video documenting how the code-base changed how the new system works
- [ ] Update main README and tutorial documentation.

ShaanGondalia:
Working PR to track ROS2 migration progress. 

Closes #350

Closes #288 (ROS2 has no ros master)

## Important Notes
Once this PR is ready for review, we will have to **extensively** test every use case locally and on the robot. Testing will likely be a month-long effort. Even after everything is tested we should choose when to merge this carefully. The entire testing workflow and onboarding process will need to be recreated, and any active development branches will have to migrate their changes to ROS2 syntax.


ShaanGondalia:
@muthuArivoli I'd like to ask for some advice on migrating [`remote_launch.py`](https://github.com/DukeRobotics/robosub-ros/blob/0b413459927205137c233bce30bb9227b8f90334/onboard/ros2_ws/src/system_utils/system_utils/remote_launch.py#L37-L39). ROS2 is pretty annoying when it comes to killing nodes. The major issue right now is that terminating the pid of a `ros2 run` command doesn't kill the nodes that were started from the command as long as there's another instance of `ros2` running. For instance:

```
  PID TTY          TIME CMD
    8 pts/0    00:00:00 entrypoint.sh
   10 pts/0    00:00:00 bash
10019 pts/10   00:00:33 python3
15351 pts/1    00:00:00 ros2
15352 pts/1    00:00:00 remote_launch
15375 pts/1    00:00:00 ros2
15376 pts/1    00:00:00 wrapper
15378 pts/1    00:00:00 data_server
15380 pts/1    00:00:00 guess_server
15382 pts/1    00:00:00 processing_serv
15413 pts/10   00:00:00 ps
```

Here, killing `15375` does not kill `wrapper`, `data_server`, `guess_server`, and `processing_serv`, all of which are nodes that were started with `ros2 service call /start_node custom_msgs/srv/StartLaunch "{package: acoustics, file: acoustics.launch.py, is_launch_file: true, args: [sim:=true]}"`.

To get around this, my naive solution is to kill all PIDs larger than the "source" `ros2` process (in this case `15375`), stopping when we reach another `ros2` process. This could also kill some processes that are unrelated, which isn't ideal. I'm thinking of creating an EOF-like process that is started after the original `ros2 run` or `ros2 launch` processes are started via a remote launch, which will signal that all of the nodes associated with the original `ros2` process are deleted. Is this a reasonable solution? I'm not super familiar with how all of these things work, so I'd like to know if you have any other thoughts.

muthuArivoli:
Instead of doing a SIGTERM, have you tried doing a SIGINT (I think use send_signal and signal.SIGINT)? I think that should work, they have different [behavior](https://github.com/ros2/launch/blob/d15f22da6cfb8be8c5374a3685f3bdc6e63cb3dc/launch/launch/launch_service.py#L208).

If it doesn't, have you considered using the python interface that ros2 provides for launching stuff itself instead of using the cmd line? The package is located here: https://github.com/ros2/launch/tree/rolling/launch. The ros2launch command eventually calls [this](https://github.com/ros2/launch_ros/blob/79f783b1fe5e06d71737e2955a8f4eb9c4f3923c/ros2launch/ros2launch/api/api.py#L141) and you can sort of mock that. So, I think you might be able to create a LaunchDescription based on the parameters you are given from the service. You can then create a LaunchService and add that description to it, and the LaunchService appears to have run and shutdown. Though they don't seem to directly support threading, so not sure about that.

If those don't work, then yeah, go with whatever works. I'll give it a bit more thought.

Also, awesome work on this ROS2 stuff, I know it can be quite a headache. Excited to see this go through!

ShaanGondalia:
> Instead of doing a SIGTERM, have you tried doing a SIGINT (I think use send_signal and signal.SIGINT)?

I recall trying this when I started a node via `ros2 run` and it didn't work. I'll give it a shot with `ros2 launch` and see if they behave properly.

> If it doesn't, have you considered using the python interface that ros2 provides for launching stuff itself instead of using the cmd line? 

I did look at `ros2launch` briefly and thought that it was viable, although to my understanding it would limit us to only running launch files. I'll look into this more if SIGINT doesn't workout.

ShaanGondalia:
Good news, SIGINT works for launch files. Unfortunately, that means that nodes ran with `ros2 run` can't get stopped remotely for now, but I've written a TODO (and eventually an issue) to track this.

vedarshshah:
ROS2 migration has been completed with `robosub-ros2`.

PR #383, opened 2022-11-19
Change controls to use ROS Actions

### LINKED ISSUE #351: Replace desired state topics with actions
## Background
Using topics for desired state in controls is an anti-pattern. Currently, the `desired_state` node expects constant publishing of a set-point (at a rate of 10Hz) to run the PID loops and produce movement. If the node doesn't receive an update on any desired state topic, it halts all movement until it receives a new message. It also halts if more than one type of control is requested in the same cycle.

This design better matches the use case of a [ROS Action](https://wiki.ros.org/actionlib), as we can model any control as an action that is requested of the controls package. The action specification includes a goal, feedback, and result. Because controls is stateless (it doesn't keep track of when a movement is complete), the feedback and result portions of the action seem useless and can be discarded. The goal can be the desired state. 

While controls doesn't need the power of feedback and result, switching to actions eliminates the need for constant publishing that can cause issues if timings are mis-aligned and simplifies logic in `desired_state.py`. Also, this should improve the process of calling controls from task planning; action calls preempt previous actions by default.

## Important Notes
Here are two designs for switching to actions:

1. Write 3 different actions for pose, velocity, and power control. This requires the creation of 3 action servers in `desired_state.py`. This is a problem because it allows for simultaneous actions (i.e. a request for velocity control wont automatically cancel a pose control action that's already running). This simplifies the calling conventions but doesn't take advantage of the built in preemption in actionlib.
1. Write 1 action that encompasses pose, velocity, and power control. The goal of this action contains the desired set-points as well as the type of control requested. This complicates the calling conventions but imposes the constraint that only one request is acted upon at any given time.

## Tasks
- [x] Write custom messages and actions to replace desired state topics
- [x] Write action servers in desired state
- [x] Update `test_state_publisher.py`
- [x] Update task planning scripts
- [x] Update pub/sub documentation in Controls README (Maybe include boilerplate for creating an action-client?)

Brokemia:
This changes controls so that instead of having to publish to specific topics at a fixed rate, we use actions to send "goals" which run until pre-empted by another goal or canceled manually. The relevant task-planning changes that come as part of this PR have been done hastily, due to our plans to refactor some of our task planning system next.

Closes #351

### DIFF
--- a/onboard/catkin_ws/src/controls/scripts/desired_state.py
+++ b/onboard/catkin_ws/src/controls/scripts/desired_state.py
@@ -68,6 +68,7 @@ class DesiredStateHandler:
         while not self.pose_server.is_preempt_requested(): 
             self.pid_manager.position_control(pose)
             rate.sleep()
+        self.pid_manager.soft_estop()
         self.pose_server.set_preempted()
 
 
@@ -90,6 +91,7 @@ class DesiredStateHandler:
         while not self.twist_server.is_preempt_requested(): 
             self.pid_manager.velocity_control(twist)
             rate.sleep()
+        self.pid_manager.soft_estop()
         self.twist_server.set_preempted()
 
     def _on_power_received(self, goal):
@@ -112,6 +114,7 @@ class DesiredStateHandler:
         while not self.power_server.is_preempt_requested(): 
             self.pid_manager.power_control(power)
             rate.sleep()
+        self.pid_manager.soft_estop()
         self.power_server.set_preempted()
 
     def _power_state_safety(self, power):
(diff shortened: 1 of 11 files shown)


Brokemia:
Verified in simulation that it doesn't crash and moves _vaguely_ in the direction given

onboard/catkin_ws/src/controls/scripts/desired_state.py
```diff
@@ -45,40 +43,77 @@ def __init__(self):
...
         """
-        self.twist = utils.parse_twist(twist)
+        self.pose_server.preempt_request = True
+        self.power_server.preempt_request = True
 
-    def _on_power_received(self, power):
+        twist = utils.parse_twist(goal.twist)
+
+        if not self._twist_state_safety(twist):
+            self.twist_server.set_aborted()
```
ShaanGondalia, line 94: Can we log an error message here

ShaanGondalia (changes requested):

onboard/catkin_ws/src/controls/scripts/desired_state.py
Brokemia: `_twist_state_safety` already logs for each axis that is out of bounds

ShaanGondalia (approved):

PR #386, opened 2022-12-04
Depthai Simulated Detection

tylerfeldman321:
- Combined `depthai_mock_camera_feed.py` and `depthai_simulate_spatial_detections.py`
- Created `depthai_simulate_detection.py`, which has functionality to run detection using the oak camera on a still image or an Image topic
- Added a launch file that takes in the feed_path, which can be either the relative path to a still image or a topic


vedarshshah:
Squash and merged locally.

PR #388, opened 2023-01-12
Fix issue with setpoint not updating in controls

Brokemia:
With the changes to controls to make it use Actions, I had missed that the setpoint needed to be continuously recalculated, as controls internally uses local poses. This is resolved now.


nickchak21 (approved):
LGTM

PR #389, opened 2023-01-14
Refactor how Task Planning uses SMACH

### LINKED ISSUE #371: Change how we use SMACH in task planning
## Background
Currently, we've used SMACH very similarly to the way we used our old task framework. Our `Task` class has been changed to extend `smach.State` and we have all our tasks extend that `Task` class. SMACH allows for entire state machines to be used as states in other state machines. These state machines obviously can't derive from `Task`, as they already need to extend `smach.StateMachine`.

Our `Task` class currently does very little. It pretty much just exists to hold CV data and otherwise works very similarly to `smach.State` itself. We should probably look into removing the `Task` class and then making our tasks extend `smach.State` and `smach.StateMachine` directly.

## Tasks
- [x] Remove `Task` class
- [x] Move CV functionality somewhere else
- [ ] Convert existing tasks to directly extend `smach.State`

Brokemia:
This refactors a large chunk of task_planning. We've extracted interfacing with Controls and CV to "interface" classes which get passed to tasks that need them. Additionally, all of our tasks now extend smach.State directly, instead of an intermediary Task class (which wasn't doing all that much anyway).

Closes #371


nickchak21 (approved):
Tested task_runner; everything works as expected. Will test and fix CV on another pull request.

PR #390, opened 2023-01-15
Pool test bugfixes

nickchak21:
This pull request fixes bugs in velocity and local position tasks. These tasks have been tested in the pool; the robot can now move in a square. The global position looks a bit wonky; right now it seems to be rotated 90 degrees relative to the local coordinate system. However, the workflow for global position tasks seems to function fine; the coordinate system is just a bit off. 

Future work will focus on making the transition between movement tasks slightly more precise. Right now the tolerance for reaching a position is pretty large, so the robot is often not in exactly the right spot as it transitions to the next movement task. Note that all tolerances and comparisons of pose are done in the global coordinate system. 

Such precision will become better once we make our robot positively buoyant again and swap in our known good PID values with all unnecessary components on the robot removed.


Brokemia (approved):
One bug fix and some tweaks to our testing code, not much could go wrong with this

Issue #392, opened 2023-01-18
RQT Fixes


vedarshshah:
# Purpose

Make a number of minor fixes and improvements to various RQT plugins.

# Tasks

- [x] Remove libGL errors that show up every time RQT is loaded.
- [x] Separate the System Usage widget into its own plugin.
- [x] Reduce the horizontal space taken by the System Usage widget.
- [x] Add a separator between launch files and Python files in the node selection dropdown of Launch Dialog.
- [x] Add help tooltips to arguments in Launch Dialog.
- [x] Add type checking to arguments in Launch Dialog.
- [x] Add regex checking to arguments in Launch Dialog.
- [x] Use checkboxes instead of text fields for boolean arguments in Launch Dialog.
- [x] Use combo boxes instead of text fields for arguments with a predefined list of possible values in Launch Dialog.
- [x] Remove QStandardPaths error that shows up every time RQT is loaded.
- [x] Add empty checking to arguments in Launch Dialog.

PR #393, opened 2023-01-21
CV RQT Fixes

vedarshshah:
Made a bunch of small changes related to RQT. See issue #392 for a complete list of changes.


landside/catkin_ws/src/gui/src/gui/launch_dialog.py
```diff
@@ -119,52 +129,153 @@ def traverse_tree(root):
...
                     default_value = arg['default']
 
                 label = QtWidgets.QLabel(arg['name'])
+
                 input = QtWidgets.QLineEdit()
                 input.setText(default_value)
 
+                toolTip = ""
+
+                if arg.get("doc") is not None:
```
ShaanGondalia, line 146: This is a pretty long if-else chain, we should consider extracting some of these conditionals to functions for more clarity.

ShaanGondalia:
We should remove `usb_camera.launch` from the launch automation script to fix the CI build failure.

vedarshshah:
> We should remove `usb_camera.launch` from the launch automation script to fix the CI build failure.

Added `rospy.spin()` to fix the issue.

landside/catkin_ws/src/gui/src/gui/launch_dialog.py
vedarshshah: Done

landside/catkin_ws/src/gui/src/gui/launch_dialog.py
```diff
@@ -87,12 +92,22 @@ def setup_node_name_box(self, selected_package):
         launchable_names = [os.path.split(f)[1] for f in launchables]
         self.node_name_box.addItems(launchable_names)
 
-    def node_name_selected(self, item_index):
+        try:
+            index_of_first_script_file = next(i for i, v in enumerate(launchable_names) if v.endswith('.py'))
```
Brokemia, line 96: I'm struggling with order of operations here, could we get some parentheses. Also, you seem to be using exceptions for control flow. That's generally a bad idea.

landside/catkin_ws/src/gui/README.md
```diff
@@ -86,7 +85,125 @@ The table will show you what nodes are running.
...
+    <td>None</td>
+  </tr>
+  <tr>
+    <td>false (no quotes, case-sensitive)</td>
+    <td>Plugin requires argument value to be a empty string</td>
+    <td>Any string</td>
+    <td>None</td>
+  </tr>
+</tbody>
+</table>
```
Brokemia, line 175: Might be good to document default values of these attributes. Or at least for `allowEmpty`, since the others are obvious

landside/catkin_ws/src/gui/src/gui/launch_dialog.py
```diff
@@ -102,49 +117,200 @@ def node_name_selected(self, item_index):
...
+
+            if arg.get("doc") is not None:
+                doc_dict = {}
+                try:
+                    doc_dict = json.loads(arg["doc"])
+                except Exception:
+                    rospy.logwarn(f"Could not parse JSON doc string for argument `{arg['name']}` in "
+                                  f"`{selected_node}`. Defaulting to unrestricted string input.")
+
+                if doc_dict.get("options") is not None:
```
Brokemia, line 146: Can't you do `"options" in doc_dict` instead of this. Same applies for a couple other places in here

landside/catkin_ws/src/gui/src/gui/launch_dialog.py
```diff
@@ -102,49 +117,200 @@ def node_name_selected(self, item_index):
...
+                    input, toolTip = self.interpret_type(selected_node, arg, doc_dict, input, toolTip)
+
+                if doc_dict.get("help") is not None:
+                    if isinstance(doc_dict["help"], str) and doc_dict["help"].strip():
+                        if toolTip:
+                            toolTip += " | Help: "
+                        toolTip += doc_dict["help"]
+
+                if doc_dict.get("allowEmpty") is not None:
+                    if type(doc_dict["allowEmpty"]) == bool:
```
Brokemia, line 162: Should you also be using `isinstance` here, since you did above for strings?

landside/catkin_ws/src/gui/src/gui/launch_dialog.py
```diff
@@ -102,49 +117,200 @@ def node_name_selected(self, item_index):
...
+
+        except Exception:
+            rospy.logwarn(f"Regex for argument `{arg['name']}` in `{selected_node}` is not valid. "
+                          f"Defaulting to unrestricted string input.")
+
+        return (input, toolTip)
+
+    def interpret_type(self, selected_node, arg, doc_dict, input, toolTip):
+        add_tooltip = True
+        if doc_dict["type"] == "bool":
```
Brokemia, line 238: A `match` statement might make more sense here than this if-else chain

landside/catkin_ws/src/gui/src/gui/launch_dialog.py
```diff
@@ -102,49 +117,200 @@ def node_name_selected(self, item_index):
...
+                elif arg["default"].lower() == "false":
+                    input = QtWidgets.QCheckBox()
+                    input.setChecked(False)
+                else:
+                    rospy.logwarn(f"Default value for argument `{arg['name']}` in `{selected_node}` is "
+                                  f"not a valid boolean. Defaulting to unrestricted string input.")
+                    add_tooltip = False
+            else:
+                input = QtWidgets.QCheckBox()
+                input.setChecked(False)
```
Brokemia, line 252:
```suggestion
            if arg.get("default") is None or arg["default"].lower() == "false":
                input = QtWidgets.QCheckBox()
                input.setChecked(False)
            elif arg["default"].lower() == "true":
                input = QtWidgets.QCheckBox()
                input.setChecked(True)
            else:
                rospy.logwarn(f"Default value for argument `{arg['name']}` in `{selected_node}` is "
                              f"not a valid boolean. Defaulting to unrestricted string input.")
                add_tooltip = False
```

landside/catkin_ws/src/gui/src/gui/launch_dialog.py
```diff
@@ -102,49 +117,200 @@ def node_name_selected(self, item_index):
...
+
+        elif doc_dict["type"] == "int":
+            int_validator = QtGui.QIntValidator()
+            if arg.get("default") is not None:
+                if (int_validator.validate(arg.get("default"), 0)[0] == QtGui.QValidator.Acceptable or
```
Brokemia, line 257: Pretty sure you can just do `arg["default"]` here instead of `arg.get("default")`, now that you've checked that the default exists. Also, I would personally check the empty string case before the valid int case, since that's easier/faster to check.

landside/catkin_ws/src/gui/src/gui/launch_dialog.py
```diff
@@ -102,49 +117,200 @@ def node_name_selected(self, item_index):
...
+                                  f"not a valid integer. Defaulting to unrestricted string input.")
+                    add_tooltip = False
+            else:
+                input.setValidator(int_validator)
+
+        elif doc_dict["type"] == "double":
+            double_validator = QtGui.QDoubleValidator()
+            if arg.get("default") is not None:
+                if (double_validator.validate(arg.get("default"), 0)[0] == QtGui.QValidator.Acceptable
+                        or arg.get("default") == ""):
```
Brokemia, line 271: Same as above

onboard/catkin_ws/src/cv/scripts/depthai_camera_connect.py
```diff
@@ -55,9 +61,18 @@ def connect(pipeline):
...
+        if rospy.is_shutdown():
+            break
+
         # Wait two seconds before trying again
         # This ensures the script does not terminate if the camera is just temporarily unavailable
-        time.sleep(2)
+        wait_secs = 2
+        start = time.time()
+        while time.time() - start < wait_secs and not rospy.is_shutdown():
+            pass
```
Brokemia, line 72:
```suggestion
        rospy.sleep(2)
```

onboard/catkin_ws/src/cv/scripts/usb_camera.py
```diff
@@ -71,7 +71,9 @@ def run(self):
                 # Sleep loop to maintain frame rate
                 loop_rate.sleep()
         except Exception:
-            rospy.loginfo("Camera not found at channel {self.channel}")
+            rospy.logerr(f"Camera not found at channel {self.channel}")
+
+        rospy.spin()
```
Brokemia, line 76: Why the spin? The only ways I can see this ending are if an error occurs. Don't you want it to stop in that case?

Brokemia (commented):
Generally looks good, these are mostly code style comments.

Also what's the purpose of the empty file at [onboard/catkin_ws/src/cv/scripts/test_images.py](https://github.com/DukeRobotics/robosub-ros/pull/393/files#diff-d320183e429808f7d54b3538bf1b81661c1581d3be2644beb1bcb4e8fef91e31)

onboard/catkin_ws/src/cv/scripts/usb_camera.py
vedarshshah: I added the spin because the CI gave an error when trying to run `usb_camera.launch` that the process exited. I can also add `usb_camera.launch` to the block list to solve the CI error. Which do you prefer?

Brokemia: I would lean towards adding it to a block list. CI isn't very helpful if it's rigged to always succeed. Maybe @nickchak21 would have a better idea on handling this though

vedarshshah:
> Generally looks good, these are mostly code style comments.
> 
> Also what's the purpose of the empty file at [onboard/catkin_ws/src/cv/scripts/test_images.py](https://github.com/DukeRobotics/robosub-ros/pull/393/files#diff-d320183e429808f7d54b3538bf1b81661c1581d3be2644beb1bcb4e8fef91e31)

`test_images.py` is not empty. You can confirm by switching to `cv-rqt-fixes` viewing the file from the Code tab above. In the diff it shows it as empty because no lines were changed; the only change to the file was that it was made executable for roslaunch to work.

landside/catkin_ws/src/gui/README.md
```diff
@@ -86,7 +85,125 @@ The table will show you what nodes are running.
...
-Data is published in percentage for CPU and GB for RAM in the System Usage box.
+#### Launch Argument Validation & Help
+
+The Launch Plugin also has the capability to validate arguments input by the user before the node is launched and provide descriptions of arguments. 
+
+To use this capability, add the `doc` attribute to `<arg>` tag in your launch file. Inside this `doc` attribute, put a JSON object.
+
+Example:
+```xml
+<arg name="count" default="" doc='{"type":"int","allowEmpty"=false,"help":"The number of messages to publish"}'>
```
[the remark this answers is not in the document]
vedarshshah: Added closing /

landside/catkin_ws/src/gui/README.md
```diff
@@ -86,7 +85,125 @@ The table will show you what nodes are running.
...
+```
+**There must NOT be anything other than the JSON object in the `doc` attribute for the validation to work.**
```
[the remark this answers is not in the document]
vedarshshah: I know that ideally we would add attributes to the XML tags to communicate additional information beyond what roslaunch supports. However, adding custom attributes to the XML makes roslaunch give warnings. For example, I added the `type` attribute to `framerate` argument in `usb_camera.launch` like so
```
<arg name="framerate" default="-1" type="int" />
```
and when launching the file the following warning printed out to the console:
```
WARNING: [/root/dev/robosub-ros/onboard/catkin_ws/src/cv/launch/usb_camera.launch] unknown <arg> attribute 'type'
```
While these are warnings and not fatal errors that stop the launch, I thought it would be quite a nuisance to have a bunch of warnings show up every time we launch a file with custom attributes in the XML. The only way  to add extra information to the `arg` tags without roslaunch printing warnings would be to add it to an attribute. The `doc` attribute is most appropriate for this as ROS does not try to interpret the `doc` attribute and so we are free to put any information there.

An additional benefit of adding information to the `doc` attribute is that it is printed out with the argument name when the `--ros-args` flag is added to the roslaunch command. Since information like `type`, `allowEmpty`, `help`, etc. are useful things to know about the arguments, it is nice to see them printed out with `--ros-args`. If we were to add this information as custom attributes, they would not be printed with `--ros-args`.

I found adding JSON objects inside the `doc` attributes to be the best way to add the information we need. I am happy to consider any alternatives you suggest.

landside/catkin_ws/src/gui/config/robosub_ros_cv_gui.perspective
```diff
@@ -4,14 +4,14 @@
...
...1afa000000000100000002fb0000005a007200710074005f0069006d006100670065005f0076006900650077005f005f0049006d0061006700650056006900650077005f005f0033005f005f0049006d006100670065005600690065007700570069006400670065007401000003c4000001dc0000019200fffffffb0000005a007200710074005f0069006d006100670065005f0076006900650077005f005f0049006d0061006700650056006900650077005f005f0037005f005f0049006d00610067006500560069006500770057006900640067006500740100000000ffffffff0000019200fffffffc0000018200000167000000c00100001afa000000000100000002fb0000005a007200710074005f0069006d006100670065005f0076006900650077005f005f0049006d0061006700650056006900650077005f005f0034005f005f0049006d006100670065005600690065007700570069006400670065007401000003c4000001dc0000019200fffffffb0000005a007200710074005f0069006d006100670065005f0076006900650077005f005f0049006d0061006700650056006900650077005f005f0038005f005f0049006d00610067006500560069006500770057006900640067006500740100000000ffffffff0000019200ffffff000005990000000000000004000000040000000800000008fc00000001000000030000000100000036004d0069006e0069006d0069007a006500640044006f0063006b00570069006400670065007400730054006f006f006c0062006100720000000000ffffffff0000000000000000')",
```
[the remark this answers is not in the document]
vedarshshah: That's just a standard part of the perspective files.  The data on that line is automatically generated by RQT; it is not something we specify.

landside/catkin_ws/src/gui/src/gui/launch_dialog.py
vedarshshah: I removed the try/except that you highlighted. I also added parentheses to the if statement on line 97. Were you suggesting to add parentheses somewhere else?

There are two other places where I use exceptions for control flow: line 138 (parsing JSON) and line 214 (validating regex). I removed the try/except on line 214 and replaced it with an if statement validating the regex.

Using a try/except to ensure JSON is valid seems to be the most accepted (maybe only) way to validate JSON in python [(see this StackOverflow post)](https://stackoverflow.com/questions/5508509/how-do-i-check-if-a-string-is-valid-json-in-python).

landside/catkin_ws/src/gui/README.md
vedarshshah: The default value for `allowEmpty` is mentioned in the bullet points under "A few more notes about the table".

For all attributes other than `allowEmpty`, if they are not specified they don't have a default value. The plugin behaves differently if an attribute is specified with any value versus if the attribute is not specified at all. Saying that the attributes have default values would imply that the plugin always behaves as if all the attributes have values, which is not true.

onboard/catkin_ws/src/cv/scripts/usb_camera.py
nickchak21: This isn't something I know much about, but I think I would prefer adding it to a block list as well. I don't generally see `rospy.spin()` used in this way.

landside/catkin_ws/src/gui/src/gui/launch_dialog.py
vedarshshah: Removed all unnecessary `.get` operations from the file. Switched the order of the conditionals in the if statement.

landside/catkin_ws/src/gui/src/gui/launch_dialog.py
vedarshshah: Addressed above

onboard/catkin_ws/src/cv/scripts/usb_camera.py
vedarshshah: Added `usb_camera.launch` to blocklist and removed spin from `usb_camera.py`.

landside/catkin_ws/src/gui/src/gui/launch_dialog.py
vedarshshah: Removed all unnecessary `.get` operations.

landside/catkin_ws/src/gui/src/gui/launch_dialog.py
vedarshshah: Replaced `type` with `isinstance`.

landside/catkin_ws/src/gui/src/gui/launch_dialog.py
vedarshshah: `match` statements were introduced in Python 3.10. The Docker container and robot use Python 3.8.10.

landside/catkin_ws/src/gui/README.md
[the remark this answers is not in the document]
Brokemia: That seems fine, I didn't know ROS would complain about unknown arguments

landside/catkin_ws/src/gui/src/gui/launch_dialog.py
Brokemia: RIP

Brokemia (approved):
Looks good

PR #397, opened 2023-02-02
Buoy task half-finished

Brokemia:
Here is about half of the buoy task, along with a little bit of testing for CV interfacing with task planning.


nickchak21 (approved):
LGTM

PR #398, opened 2023-02-04
Compose files for specific robots

AndrewCouncil:
Created docker compose file for specific robots and a line to the entrypoint to export an environment variable from the compose file

### DIFF
--- a/docker-compose-cthulhu.yml
+++ b/docker-compose-cthulhu.yml
@@ -1,0 +1,47 @@
+version: "3.3"
+
+services:
+  onboard-cthulhu:
+    image: dukerobotics/robosub-ros:onboard
+    container_name: onboard-cthulhu
+    environment:
+      - ROBOT_NAME=cthulhu
+    privileged: true
+    ports: 
+      - 2200:2200
+    networks:
+      local_net:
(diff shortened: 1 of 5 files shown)


docker-compose-cthulhu.yml
```diff
@@ -0,0 +1,47 @@
+version: "3.3"
+
+services:
+  onboard-cthulhu:
+    image: dukerobotics/robosub-ros:onboard
+    container_name: onboard-cthulhu
+    environment:
+      - ROBOT_NAME=cthulhu
```
ShaanGondalia, line 8: Is the only difference in these compose files the existence of these environment variables? If so is there some other way to inject the robot name into the container without needing duplicate compose files? Maybe we can refactor our code to use [Extension](https://docs.docker.com/compose/extends/#extending-services).

Brokemia: Made that change

AndrewCouncil:
this has now been tested and is working. for any shell open under root user the environment variable ROBOT_NAME will be set to the name of the robot. can be checked in python using [os.environ](https://docs.python.org/3/library/os.html#os.environ)

Brokemia (approved):
Looks good

PR #402, opened 2023-02-16
CV Compressed Images

vedarshshah:
All publishers in the `cv` package were transitioned to use the `CompressedImage` message format instead of the raw `Image` format. This was done to use less memory and improve speed.

Additionally, all subscribers in the `cv` package were transitioned to dynamically get the type (`Image` or `CompressedImage`) of image topics and handle both types. Moving forward, most image topics that will be subscribed to will use `CompressedImage`, however bag files recorded before this transition still use `Image`, so the dynamic type capability was added to support both.

Waiting to merge #399 before marking this as ready for review.


onboard/catkin_ws/src/cv/README.md
```diff
@@ -161,6 +161,14 @@ To simulate camera feed and then run a model on the feed from the left camera. W
...
-`roslaunch cv test_images.launch feed_path:=../assets/buoy.jpg topic:=/camera/right/image_raw framerate:=30`: Publishes the still image buoy.jpg in cv/assets to the topic '/camera/right/image_raw' 30 times per second.
+`roslaunch cv test_images.launch feed_path:=../assets/buoy.jpg topic:=/camera/right/compressed framerate:=30`: Publishes the still image buoy.jpg in cv/assets to the topic '/camera/right/compressed' 30 times per second.
+
+# Utils
+
+The `utils.py` file has two classes that are used by many other files in this package.
+
+`DetectionVisualizer` provides functions to draw bounding boxes and their labels onto images. It is used by DepthAI files when publishing visualized detections.
+
+`ImageTools` provides functions to convert between OpenCV, ROS Image, and ROS CompressedImage formats. All scripts in this package use `ImageTools` to perform converstions between these types. `cv_bridge` is not used by any file or class in this package other than `ImageTools` itself.
```
Brokemia, line 174:
```suggestion
`ImageTools` provides functions to convert between OpenCV, ROS Image, and ROS CompressedImage formats. All scripts in this package use `ImageTools` to perform conversions between these types. `cv_bridge` is not used by any file or class in this package other than `ImageTools` itself.
```

onboard/catkin_ws/src/cv/scripts/test_images.py
```diff
@@ -69,10 +69,9 @@ def run_still(self):
         """Publish a still image to topic with the specified framerate."""
 
         image = cv2.imread(self.feed_path, cv2.IMREAD_COLOR)
-        image_msg = self.cv_bridge.cv2_to_imgmsg(image, 'bgr8')
         loop_rate = rospy.Rate(self.framerate)
-
         while not rospy.is_shutdown():
+            image_msg = self.image_tools.convert_to_ros_compressed_msg(image)
             self.image_publisher.publish(image_msg)
```
Brokemia, line 75: Why is the conversion being done in the loop now? Is there a possibility for `image` to change?

onboard/catkin_ws/src/cv/scripts/utils.py
```diff
@@ -53,3 +54,298 @@ def visualize_detections(self, frame, detections):
             self.rectangle(frame_copy, bbox)
 
         return frame_copy
+
+
+class ImageTools(object):
```
Brokemia, line 59: Can this be its own file, rather than throwing it in `utils.py`?

onboard/catkin_ws/src/cv/scripts/utils.py
```diff
@@ -53,3 +54,298 @@ def visualize_detections(self, frame, detections):
...
+                                          compressed_format='jpg'):
+        """
+        Convert from cv2 image to ROS CompressedImage.
+        """
+        return self._cv_bridge.cv2_to_compressed_imgmsg(cv2_data,
+                                                        dst_format=compressed_format)
+
+    def convert_ros_msg_to_ros_compressed_msg(self, image,
+                                              image_encoding='bgr8',
+                                              compressed_format="jpg"):
```
Brokemia, line 104: Inconsistent quotes between ' or ".
I generally don't care, but these are right next to each other

onboard/catkin_ws/src/cv/scripts/utils.py
```diff
@@ -53,3 +54,298 @@ def visualize_detections(self, frame, detections):
...
+        Convert any kind of image to cv2.
+        """
+        cv2_img = None
+        if type(image) == np.ndarray:
+            cv2_img = image
+        elif image._type == 'sensor_msgs/Image':
+            cv2_img = self.convert_ros_msg_to_cv2(image)
+        elif image._type == 'sensor_msgs/CompressedImage':
+            cv2_img = self.convert_ros_compressed_to_cv2(image)
+        else:
```
Brokemia, line 125: This method (and the other ones like it) seems to assume the image encoding is "bgr8", is that a good idea?

onboard/catkin_ws/src/cv/scripts/utils.py
```diff
@@ -53,3 +54,298 @@ def visualize_detections(self, frame, detections):
...
+        return cimg_msg
+
+    def convert_to_cv2(self, image):
+        """
+        Convert any kind of image to cv2.
+        """
+        cv2_img = None
+        if type(image) == np.ndarray:
+            cv2_img = image
+        elif image._type == 'sensor_msgs/Image':
```
Brokemia, line 121: Why are we doing this test rather than `type(image) == Image` (with Image imported of course)

onboard/catkin_ws/src/cv/scripts/utils.py
```diff
@@ -53,3 +54,298 @@ def visualize_detections(self, frame, detections):
...
+        return ros_cmp
+
+    def convert_depth_to_cv2(self, image):
+        cv2_img = None
+        if type(image) == np.ndarray:
+            cv2_img = image
+        elif image._type == 'sensor_msgs/Image':
+            image.encoding = 'mono16'
+
+            encoding = image.encoding.split(";")[0]
```
Brokemia, line 214: You set image.encoding right before this, so this will always result in "mono16"

Brokemia (commented):
@vedarshshah

onboard/catkin_ws/src/cv/scripts/test_images.py
vedarshshah: Put conversion outside loop.

onboard/catkin_ws/src/cv/scripts/utils.py
vedarshshah: Removed assignment to image.encoding.

onboard/catkin_ws/src/cv/scripts/utils.py
vedarshshah: Changed all type tests to use `type` function.

onboard/catkin_ws/src/cv/scripts/utils.py
vedarshshah: Moved `ImageTools` class to `image_tools.py` and changed imports for class in other files.

onboard/catkin_ws/src/cv/scripts/utils.py
vedarshshah: Added parameters to set image encoding, with "bgr8" as default.

vedarshshah:
@Brokemia Letting you know that all comments have been addressed and everything has been tested and works. Please take a look and approve.

Brokemia (approved):
Looks good

PR #405, opened 2023-02-23
removed nested loop from postion_pid.launch

alexpool1:
Switched back to one loop controls so that it is easier to tune on the new robot. Will add commit for changing comments and README.


Brokemia:
Our pool test today was done using the old double loop, and the new robot seemed to work just fine. Probably not going to merge this for now, since everything worked

PR #411, opened 2023-03-04
Spatial Detection Performance Improvements

### LINKED ISSUE #406: Spatial Detection Performance Improvements
# Background

Currently, our DepthAI spatial detection pipeline produces detections at 1Hz. The detection rate must be improved for the robot to effectively navigate to the buoys and other targets.

# Purpose

To improve the detection rate of the DepthAI spatial detection pipeline. We wish to get the detection rate to at least 15 Hz and hopefully to 30 Hz.

# Tasks
- [x] Remove the `loop_rate` in `depthai_spatial_detection.py` that artificially limits the detection rate to 1Hz.
- [x] Parametrize all output queues except "detections" so they can be turned off and on by changing values of arguments in the launch file. Set the default to `True` for the "rgb" queue, `False` for all other queues.
- [x] Publish the depth map to a topic if its output queue exists.
- [x] Parametrize the `sync_nn` argument to the `build_pipeline` function so its value is controlled by the launch arguments. Set the default to `True`.
- [x] Benchmark the detection rate after all changes are made above. Obtain two benchmarks: one with the "rgb" output queue turned on (and all other output queues turned off) and one with all output queues (including "rgb") turned off. Both benchmarks must be obtained on the robot.

vedarshshah:
Improved speed of spatial detection using DepthAI from 1Hz to 22Hz. Closes #406. Will be ready for review after #402 is merged in.


Brokemia (approved):
I don't understand CV stuff well enough to pass effective judgement, but nothing stands out as obviously wrong

onboard/catkin_ws/src/cv/scripts/depthai_simulate_detection.py
```diff
@@ -213,16 +216,11 @@ def _run_detection_on_image_topic(self, device):
         """
         TopicType, _, _ = rostopic.get_topic_class(self.feed_path)
         rospy.Subscriber(self.feed_path, TopicType, self._update_latest_img)
-        loop_rate = rospy.Rate(1)
```
ShaanGondalia, line 216: Is there a particular reason for not limiting the loop frequency?

vedarshshah: Back when we developed the first DepthAI script, we needed to limit the loop frequency for it to work properly. However, later versions of DepthAI scripts, including the current versions, do not require a limit on the loop frequency to function. In fact, we discovered the limit on the loop frequency as significantly reducing the detection rate, which we needed to improve. Therefore, we got rid of the limits on the loop frequency.

Issue #410, opened 2023-02-25
CV Sonar Full Merge


ricohasgithub:
### Purpose

Currently, both cv (using stereo vision) and sonar each have their own methods of obtaining depth information given a target. This update aims to integrate the cv and sonar code bases together to create a unified way of detecting the position of detected objects in water, specifically by using sonar to complement cv's current abilities. In addition to the full sonar library, this merge will mainly concern the script ```depthai_sonar_spatial_detection.py``` and the respective launch file. The script ```depthai_sonar_spatial_detection.py``` will run the regular depthai cv detection code, and once an object comes into frame, it will make a request to the sonar library.

vedarshshah:
Merged. Sonar code was put into `depthai_spatial_detection.py` and `depthai_sonar_spatial_detection.py` was deleted.

PR #412, opened 2023-03-04
CV + Sonar Full Integration

### LINKED ISSUE #400: Sonar Fixes
# Background

The sonar uses the blue robotics ping-360 library to interact with the physical sonar device, however, the sonar code which we use to connect to the sonar has a few issues. The original system which we developed to communicate with the sonar through ros actions was too slow for the quick sonar pings which we need for sonar-CV integration. Now we are switching to the usual rosnode system but there are lots of artifacts currently within the code base. 

The current sonar code is functional but it needs changes before it is ready to be marked as its own package.

# Purpose

These changes will make the sonar system more robust as currently there are a few problems which will hinder the modularity of the system. In the sonar setup, the sonar is currently found through a hardcoded port. This is a problem as the same code will not work on multiple robots or within a different system. The 'SERIAL_PORT_NAME' should be changed to be found through an automated system similar to 'imu.py'. 

The sonar does not run off of a range, rather it has a 'duration', 'number of samples', and 'period' which it uses to get sonar data over a range of points. The duration is the length of one ping and the period is how often the sonar pings. These two together control the effective range of the sonar. The number of samples is the number of pings used throughout the range determined by the other two parameters. Lowering this will make the sonar faster but more granular. The current way these parameters are calculated is through a linear function, however, they should be changed to match the original code in the pingviewer application made by bluerobotics. 

The data we get from the sonar is stored in an array where the columns are the number of samples and the rows are the angles. Visualizing with numpy, the sonar image that is formed is a polar image in a rectangular format. This is confusing but it works for most purposes. For better CV image processing of the gate, however, it should be changed to polar format.

The first .75 meters of the sonar scan are also noise so they should be dynamically filtered out.

# Tasks
- [ ] Find 'SERIAL_PORT_NAME' dynamically
- [ ] Find a better system to determine the period and duration
- [ ] Create a polar image processing function in 'sonar_image_processing'
- [ ] Create a dynamically filtering method so the first few points of the sonar are filtered out
- [ ] Prune extraneous code from the sonar such as the rosaction code (once the node works)
- [ ] Make a new custom message for the sonar to communicate to CV and task planning over. 
- [ ] Need to test the 'to_robot_position' function to make sure that the position taken from the polar image is the same as the physical location of the object.
- [ ] Update the 'get_xy_of_object_in_sweep' method to be more robust for finding a small cluster of points (right now it takes the average which is prone to extraneous values).
- [ ] The sonar is currently started called from motion.launch. If the thrusters aren't turned on then the sonar isn't either as it is connected to the thrusters power. Also - there are issues if the thrusters are not turned on before the docker container is made! The docker needs to be able to see the sonar in the port for it to be forwarded to the container. Eventually, the sonar will be connected to the main powerbus without the need to turn on the thruster but for now the code should not error out if the sonar is not available.
  - [ ] In the 'sonar_publishing' file, it should throw an error but not stop the program from running. 
  - [ ] In the same file, whenever the sonar is called and disabled it should return an error message.
  - [ ] All other files which call the sonar should be able to interperet this error
- [ ] Remove the 'test_scan_and_finding_gate_posts' method from the sonar class (move it somewhere else)
- [ ] Remove the 'test_gate_from_npy_file' and 'test_buoy_from_npy_file' methods as well.
- [ ] In sonar image processing, remake the gate finding method to use a more robust system for finding the gate. Currently it detects all "circular" objects by perimeter and the finds the one which is the most concave. This should be switched to a different system which can accurately find the gate.
- [ ] The node system needs to be setup for task-planning to interact with the gate finding method.
- [ ] If task-planning is still in the gate stage, the CV method should not be able to be called until the gate is found as that could cause some errors finding the gate.
- [ ] in 'sonar_image_processing' remove the find_buoy' method (or change it)
- [ ] Add buoy hit detection
- [ ] Fix FOV camera buoy correlation

ricohasgithub:
Full CV and Sonar integration. Resolves #400 and #410, and code has been tested in two pool tests to ensure correctness; contains the entirety of the sonar branch which had not yet been merged into master so far.


onboard/catkin_ws/src/cv/scripts/depthai_sonar_spatial_detection.py
```diff
@@ -0,0 +1,463 @@
...
+
+        self.sonar_requests_publisher = rospy.Publisher(
+            SONAR_REQUESTS_PATH, sweepGoal, queue_size=10)
+        self.sonar_response_subscriber = rospy.Subscriber(
+            SONAR_RESPONSES_PATH, sweepResult, self.update_sonar)
+
+        self.pose_publisher = rospy.Publisher(POSE_PUBLISHER_PATH,
+                                              Pose, queue_size=10)
+
+    def build_pipeline(self, nn_blob_path, sync_nn=True):
```
Brokemia, line 68: Is this stuff copy and pasted from the non-sonar DepthAI code? If so, could we make this extend the other class, or at least extract the common code somewhere else? 
I was told to tag @ricohasgithub

onboard/catkin_ws/src/cv/scripts/depthai_sonar_spatial_detection.py
```diff
@@ -0,0 +1,463 @@
...
+            sonar_request_msg = sweepGoal()
+            sonar_request_msg.start_angle = left_end
+            sonar_request_msg.end_angle = right_end
+            sonar_request_msg.distance_of_scan = SONAR_DEPTH
+            self.sonar_requests_publisher.publish(sonar_request_msg)
+
+            # Try calling sonar on detected bounding box
+            # if sonar responds, then override existing robot-frame x, y info;
+            # else, keep default
+            if not (self.sonar_response == (0, 0)) and self.in_sonar_range:
```
Brokemia, line 305: Is there a better way to indicate errors than 0,0

onboard/catkin_ws/src/cv/scripts/depthai_sonar_spatial_detection.py
```diff
@@ -0,0 +1,463 @@
...
+                det_coords_robot_mm = (self.sonar_response[0],
+                                       self.sonar_response[1], y_cam_meters)
+
+                # Publish Pose to task planning directly; based on
+                # latest sonar update
+                published_pose = Pose()
```
Brokemia, line 311: I feel like a Point is better if we aren't using the orientation to store any information

onboard/catkin_ws/src/cv/scripts/depthai_sonar_spatial_detection.py
```diff
@@ -0,0 +1,463 @@
...
+                published_pose.position.z = det_coords_robot_mm[2]
+                published_pose.orientation.x = 0
+                published_pose.orientation.y = 0
+                published_pose.orientation.z = 0
+                published_pose.orientation.w = 1
+
+            self.pose_publisher.publish()
+            self.publish_prediction(
+                bbox, det_coords_robot_mm, label, confidence,
+                (height, width), self.in_sonar_range)
```
Brokemia, line 323: Even if sonar gives an error by returning (0, 0), this might still think it's using sonar

onboard/catkin_ws/src/cv/scripts/depthai_sonar_spatial_detection.py
```diff
@@ -0,0 +1,463 @@
...
+        published_pose = Pose()
+        published_pose.position.x = object_msg.coords.x
+        published_pose.position.y = object_msg.coords.y
+        published_pose.position.z = object_msg.coords.z
+        published_pose.orientation.x = 0
+        published_pose.orientation.y = 0
+        published_pose.orientation.z = 0
+        published_pose.orientation.w = 1
+
+        self.pose_publisher.publish()
```
Brokemia, line 368: This is already being done in the above method

onboard/catkin_ws/src/cv/scripts/depthai_sonar_spatial_detection.py
```diff
@@ -0,0 +1,463 @@
...
+        return True
+
+    def update_sonar(self, sonar_results):
+        """
+        Callback function for listenting to sonar response
+        Updates instance variable self.sonar_response based on
+        what sonar throws back if it is in range (> 1.5m)
+        """
+        # Check to see if the sonar is in range - are results from sonar valid?
+        if sonar_results.y_pos > 1.75:
```
Brokemia, line 400: Put 1.5 or 1.75 as a constant somewhere

onboard/catkin_ws/src/sonar/README.md
```diff
@@ -0,0 +1,2 @@
+# Sonar
```
Brokemia, line 1: image

onboard/catkin_ws/src/sonar/scripts/sonar.py
```diff
@@ -0,0 +1,285 @@
...
+from tf import TransformListener
+
+
+class Sonar:
+    """
+    Class to interface with the Sonar device.
+    """
+
+    # only for local testing with the sonar script
+    IF_LOCAL_TEST = True
```
Brokemia, line 19: Is this something that needs to be configurable?

onboard/catkin_ws/src/sonar/scripts/sonar.py
```diff
@@ -0,0 +1,285 @@
...
+
+    # PORT of the salea is ttyUSB2 for testing
+    SERIAL_PORT_NAME = "/dev/ttyUSB2"
```
Brokemia, line 22: Also configurable? Maybe with the same method as finding the DVL and IMU?

onboard/catkin_ws/src/sonar/scripts/sonar.py
```diff
@@ -0,0 +1,285 @@
...
+
+    def get_sweep(self, range_start, range_end):
+        """ Get data along a range of angles
+
+        Args:
+            range_start (int, optional): Angle to start sweep in gradians.
+                    Defaults to 100.
+            range_end (int, optional): Angle to end sweep in gradians.
+                    Defaults to 300.
+
```
Brokemia, line 104: Doesn't default to anything but says it does

onboard/catkin_ws/src/sonar/scripts/sonar.py
```diff
@@ -0,0 +1,285 @@
...
+        pose = self.to_robot_position(center_angle, mean_index)
+        return pose.position.x, pose.position.y
+
+    def get_max_bytes_along_sweep(self, start_angle, end_angle):
+        """ Execute a sweep and get the largest activation for each angle
+
+        Args:
+            start_angle (int): _description_
+            end_angle (int): _description_
+
```
Brokemia, line 170: \_description\_

onboard/catkin_ws/src/sonar/scripts/sonar.py
```diff
@@ -0,0 +1,285 @@
...
+        pos_of_point.orientation.z = 0
+        pos_of_point.orientation.w = 1
+
+        transformed_pose = sonar_utils.transform_pose(
+            self.listener, pos_of_point)
+
+        return transformed_pose
+
+
+def test_scan_and_finding_gate_posts():
```
Brokemia, line 231: Could tests get their own file?

onboard/catkin_ws/src/sonar/scripts/sonar.py
```diff
@@ -0,0 +1,285 @@
...
+    BAUD_RATE = 2000000
+
+    sweep_data = sonar.sweep_biggest_byte(100, 300)  #180deg in front
+    print(f"Distance to object: {sonar.get_distance_of_sample(
+                                sweep_data[0])} | Angle: {sweep_data[2]}")
+
+    FOR STARTING A WEB SERVER IN FOLDER::: RUN "python -m http.server 8000"
+    test_buoy_from_npy_file(os.path.join(os.path.dirname(__file__),
+    'sampleData', 'gate.npy'))
+"""
```
Brokemia, line 285: Might be good README material

onboard/catkin_ws/src/sonar/scripts/sonar_image_processing.py
```diff
@@ -0,0 +1,283 @@
...
+    assert filename.endswith('.bin'), 'filename must be a .bin file'
+
+    parser = get_bin_file_parser(filename)
+
+    data_list = []
+    for index, (timestamp, decoded_message) in enumerate(parser):
+        if index >= start_index and index <= end_index:
+            data_list.append(decoded_message.data)
+
+    JPEG_SAVE_PATH = os.path.join(os.path.dirname(__file__),
```
Brokemia, line 51: Why is this all caps, it looks like a constant

onboard/catkin_ws/src/sonar/scripts/sonar_publisher.py
```diff
@@ -0,0 +1,59 @@
+
+import rospy
+from sonar import Sonar
+from custom_msgs.msg import sweepResult, sweepGoal
+from sonar_utils import degrees_to_centered_gradians
+from sonar_image_processing import scan_and_build_sonar_image
+
+
+DEBUG_MODE = True
```
Brokemia, line 9: Should this be configurable?

Brokemia (changes requested):
Don't forget to remove the random images

onboard/catkin_ws/src/cv/scripts/depthai_sonar_spatial_detection.py
ricohasgithub: That's a good point. I think the code is slightly different as we took out some stuff with stereo that we won't need due to the use of sonar to help speed things up. We can definitely consider this in a future update.

onboard/catkin_ws/src/cv/scripts/depthai_sonar_spatial_detection.py
ricohasgithub: Currently in_sonar_range will only ever be true if sonar is running and in range, so the only case for if sonar_response is (0, 0) is when the sonar has not responded, which in case in_sonar_range will be False.

Brokemia (approved):
Looks good now

vedarshshah:
Please hold off on merging this until #411 and #402 have been merged in as their changes must be merged into the CV portion of this branch.

PR #413, opened 2023-03-23
CV Video Encoding

### LINKED ISSUE #401: DepthAI Video Encoding (DRAFT)
# Background

Currently, to record footage using the DepthAI camera we use `depthai_publish_image_stream.py`. It obtains still images from the stereo camera and publishes them to a topic. We then `rosbag` the topic to record the image stream.

`depthai_publish_image_stream.py` publishes images at only a few Hz. It is very slow and leads to poor quality footage. 

Instead of obtaining still images and publishing them to a topic, we can instead use the [DepthAI VideoEncoder](https://docs.luxonis.com/projects/api/en/latest/components/nodes/video_encoder/) node. This node encodes the image stream from the camera in H.264 or H.265 and sends the encoded stream to the host computer. This encoded stream can be saved to a file on the host computer and then converted into a playable video file using `ffmpeg`.

Thus, VideoEncoder enables us to record footage from the camera at a higher frame rate. However, the output of VideoEncoder is encoded and thus cannot be viewed live, unlike publishing images to a topic. The output must be converted to a video file before it can be viewed.

# Purpose
Develop a script using the DepthAI VideoEncoder node to obtain H.265 encoded video from the RGB and mono cameras.

# Tasks
TODO

vedarshshah:
Added capability to encode the RGB, mono left, mono right, disparity, and depth streams in H264/5 format, save the encoded streams to a file, and convert the encoded streams into playable video files. Closes #401.


Brokemia (approved):
LGTM

PR #418, opened 2023-04-06
Camera relay hard reset

mbryant2025:
Adds functionality that toggles on/off the power on the POE-enabled camera. See onboard/catkin_ws/src/offboard_comms/README.md for primary documentation.


Brokemia (approved):
A few changes that probably should've come with a different PR, but acceptable

PR #421, opened 2023-04-14
CV RQT Launch Plugin Improvements

### LINKED ISSUE #403: RQT Launch Plugin Improvements
# Purpose
To make a variety of improvements to the RQT Launch Plugin. See the task list for more details.

# Tasks
- [x] Modify `remote_launch.py` in the `system_utils` package on onboard so that if a node launched from `remote_launch` is terminated, a function within `remote_launch` is called.
- [x] Modify `remote_launch.py` so that it publishes a message to a topic whenever it launches or terminates a node, and also when a node is terminated on its own (use the function developed in the previous task). The message should contain the PID, package, file, and launch args that were used when the node was launched.
- [x] Modify `launch_widget.py` in the `gui` package on landside so that whoever a node is terminated on its own that was originally launched by the plugin, it removes the node from the table.
- [x] Add a configuration dialog to Launch Plugin ([see the `trigger_configuration` function on this page](http://wiki.ros.org/rqt/Tutorials/Writing%20a%20Python%20Plugin)). In the configuration dialog, there should be one checkbox whose value should be sent to Launch Widget (the default value is False). When the checkbox is checked (value is True), the plugin should display all nodes launched by remote_launch (not just nodes launched by the plugin) and allow the user to terminate any node (including nodes not launched by the plugin) that is displayed in the table. This means that if there is more than one instance of the Launch Plugin running and being used to launch nodes, then all instances will display all nodes that have been launched by all instances and they all can terminate each others' node as well.
- [x] Make the part of Launch Dialog that displays the arg inputs scrollable. If the arguments for any launch file take up less space than 6 arguments with QLineEdit inputs, then no scroll bar should show up. If the arguments for any launch file take up more space than that, then a scroll bar should show up. Do not include the Package, Node Name, or Other Args inputs in the scrollable area.
- [x] Remove the Args column from the table in Launch Widget.
- [x] For arg tags with defined `value`s, do not display an input dialog. Simply display the set value next to the argument name. Do not pass a value for the argument when calling remote launch.
- [x] By 4/12 - When the user double clicks on any row in the table (except the header row) a dialog should show up that displays the PID, Package Name, File Name, and Args of the node. The Args should be displayed in a table with two columns: the argument name and value. There should be a single "Close" button to close the dialog.
- [x] By 4/12 - In the Package Name and Node Name boxes in Launch Dialog, display the packages/files alphabetically by file name. All launch files should be displayed before Python files, and each group of files should be sorted alphabetically. Depending on how your computer is configured, the packages/files may already be sorted alphabetically, but the is not the case on the robot, so we need to write code to ensure it is always alphabetical.
- [x] By 4/12 - In the configuration dialog, add an option that, when checked, does not remove nodes that were terminated on their own or by another launch plugin from the table; nodes should only be removed from the table when the user selects a row and presses Fn+Delete. The row can either be of a running node, in which case the node is terminated before being removed from the table, or the row can be of an already terminated node, in which case simply delete row from the table.
- [x] By 4/12 - In the configuration dialog, add an option that, when checked, adds a column to the table that indicates whether the node is currently running or has been terminated.
- [x] By 4/15 - Make the stop node operation asynchronous. See `camera_status_widget.py` for an example on how to use `QThreadPool` for asynchronous operations.
- [x] By 4/15 - When the user presses Fn+Delete to stop a node but the operation is not successful, display a dialog notifying the user that the node could not be terminated.

gummydumpling-2001:
CV made a variety of improvements to the CV Launch Plugin in the CV GUI. Closes #403.


Brokemia (approved):
Not sure I'm qualified to review this, but looks okay

PR #423, opened 2023-04-15
RQT Topic Plugin

### LINKED ISSUE #419: RQT Topic Plugin
# Background

When performing object detection, CV publishes CVObject messages containing the bounding box, class, and (x, y, z) coordinates of the detected object. Task planning then moves the robot to those coordinates. Therefore, the coordinates are the most important part of the CVObject messages.

On the other hand, we currently do not have a way to view from our GUI the live coordinates that are being published. We must use the `rostopic` command in a terminal window to view the latest messages being published.

Since the coordinates are the most important part of the message, we must have a way to view them from our GUI, alongside the live camera feed, live bounding boxes, and the other features of our GUI.

There exists a plugin called `rqt_topic` that provides this functionality. It allows one to view all active topics, subscribe to a selected group, and view the latest message published to the topic in real time.

# Purpose

To add the `rqt_topic` plugin to the CV GUI for viewing coordinates published by CV live from the GUI.

# Tasks

Complete by 4/15.

- [x] Add the installation command for `rqt_topic` to the landside Dockerfile.
- [x] Incorporate the `rqt_topic` plugin into the CV GUI perspective such that it can be viewed simultaneously with at least two live camera feeds and with the launch plugin.

maxwellmlin:
Allows CV to see the latest messages being published in our GUI.

- Adds ros-noetic-rqt-topic to the landside Dockerfile.
- Incorporates the Topic Monitor plugin into the CV GUI perspective in a tab with the bottom right Image View.

Closes #419.


Brokemia (approved):
I really dislike these impossible to read perspective files, but I don't think that's our fault, so looks good

PR #425, opened 2023-04-20
Camera hard reset relay configuration change

mbryant2025:
Responds to changing the wiring of the relay to enable and disable the camera. Relay is now in normally closed state and only requires power to disable. Additionally, the publishing rate of the associated topics were decreased in response to CV GUI requirements.


Brokemia:
Merged into pool-test-apr-22, so this PR is no longer required

PR #442, opened 2023-09-16
sonar fixes

Will-Denton:
Sonar fix so it doesnt error


vedarshshah (approved):
I don't understand this but LGTM!

PR #459, opened 2023-09-29
System Info Publisher

vedarshshah:
Fixed a bug in system info publisher that caused it to crash when run locally (not on robot). Added system info publisher to motion.launch.


Brokemia (approved):
Seems fine

PR #463, opened 2023-10-05
Sensor Check Automation

### LINKED ISSUE #437: Sensor Check Automation
# Background

The mechanical and electrical teams will meet outside of regular meeting ours prior to pool tests to prepare the robot for the pool tests. For the robot to work as intended, the computer needs to receive data from all sensors on the robot and be able to spin the thrusters. Currently, these checks would need to be performed by a member of the software team as the mechanical and electrical members do not have the requisite software knowledge. Thus, a software team member would need to be present in the preparation meeting.

Even if a software team member is present to perform these checks, running all scripts and performing all checks is time consuming, and there is a chance that something is not checked.

# Purpose

Develop a script that automatically checks that all robot sensors are working as expected and that the thrusters can be commanded to move.

# Tasks
- [x] Check if data is received from the DVL.
- [x] Check if data is received from the IMU.
- [x] Check if data is received from the pressure sensor.
- [x] Check if the state is being published.
- [x] Check if a camera feed is being published from CV.
- [x] Check if data is received from sonar.
- [x] Check if the thrusters can be moved. Warn the user of the script that the thrusters will now be moved, and proceed only when the user has confirmed it is save to do so. Test each thruster individually, telling the user which thruster should be moving. Ask the user to confirm the specified thruster is moving before moving to the next thruster.

# Notes

This script should be built to publish the status of all the sensors to a topic, which can be leveraged by the future GUI to display the sensors status. This script should do this in addition to printing outputs to the terminal.

This script will need to be updated as the year progresses based on real-world feedback from the pool tests and also since we will be adding new sensors and rewriting large portions of our software.

ricohasgithub:
This PR will merge `sensor-check` into `master`. Closes #437 ; this addresses all of the sensors listed with the ability to test thrusters. To run the automated sensor check script, run `roslaunch system_utils sensor_check.launch`. The script will prompt whether to test the sensors, and upon response from the user via the terminal, will proceed with the tests.


vedarshshah (approved):
LGTM!

PR #476, opened 2023-10-25
Fix Deprecated DepthAI Code

vedarshshah:
DepthAI deprecated `CameraBoardSocket.RGB`, `.LEFT`, and `.RIGHT`, which were used in `depthai_publish_save_streams.py` and `depthai_spatial_detection.py`. They were changed to `.CAM_A`, `.CAM_B`, and `.CAM_C` respectively.


ricohasgithub (approved):
Looks good to me!

PR #481, opened 2023-11-04
Fixes from Oct 28 Pool Test

vedarshshah:
- Added `ROBOT_NAME=oogway` env variable to `docker run` command for onboard.
- Fixed bug and removed y offset for buoy task in `comp_2023.py`.
- Updated FTDI string for pressure sensor.
- Modified DVL config and code to swap X and Y axes and stop negating Y axis.
- Added check to pressure sensor code to not publish depth when is greater than 7 meters.
- Fixed `dual_upload.sh`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ Use these instructions when running code on the robot itself.
     ```
 1. There, make sure the latest code from this repo is pulled. Then, run the onboard container if it's not already running.
     ```bash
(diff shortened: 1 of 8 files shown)


vedarshshah (commented):
L I can edit ur comments

Will-Denton (approved):

PR #483, opened 2023-11-10
Thruster Speeds Panel

### LINKED ISSUE #451: Thruster Speeds Panel
## Motivation
Our old rqt controls GUI had a plugin for publishing speed commands for Oogway's 8 thrusters. We need to transition this plugin to our Foxglove GUI.

## Requirements
- [ ] Create a panel to set thruster speeds on the robot:
   - [ ] There should be 8 different text fields, each appropriately labeled with the correct thruster position. These labels and text fields should be displayed in a `4 x 2` grid (see [Grid v2](https://mui.com/material-ui/react-grid2/)).
   - [ ] The inputs into each field should be constrained to an `int` between -127 and 128. You can look into the Validation section in [Material UI Text Field](https://mui.com/material-ui/react-text-field/) on how to do that.
- [ ] Create another panel to view the current thruster speeds on the robot:
   - [ ] There should be 8 different text boxes, each appropriately labeled with the correct thruster position. These labels and text boxes should be displayed in a `4 x 2` grid (see [Grid](https://mui.com/system/grid/)).
- [ ] (FOR LATER) Display warning if another node is currently publishing to `/offboard/thruster_speeds`[^1]

## Notes
The `thruster-speeds-panel` has been forked off of `publish-topic-panel`. This means a lot of code is extraneous and will be deleted in the final product.

## UI
Consider using [Material UI Text Field](https://mui.com/material-ui/react-text-field/) as the input component. To import MUI, look at the examples in the Material UI Text Field page.

## Implementation
Thruster speeds are sent using a publisher. The topic for that is `/offboard/thruster_speeds`, which uses the message type `custom_msgs/ThrusterSpeeds`.

For instance, to run all the thrusters at speed 10, you can use:
```
rostopic pub -r 10 /offboard/thruster_speeds custom_msgs/ThrusterSpeeds '{speeds: [10,10,10,10,10,10,10,10]}'
```
Note that thruster speeds are sent at a rate of 10 Hz.

In order to get the current publisher speeds, create a subscriber to correct topic and message type. Extract the correct speed value for each thruster, and populate the relevant text boxes.

### Thruster configurations for Oogway
The 8 `int` values in the `speeds` array corresponds to the following thrusters on the robot:
- `speeds[0]: bottom_front_left`
- `speeds[1]: front_left`
- `speeds[2]: front_right`
- `speeds[3]: bottom_front_right`
- `speeds[4]: bottom_back_left`
- `speeds[5]: back_left`
- `speeds[6]: bottom_back_right`
- `speeds[7]: back_right`

You can read more about the thruster configs on Oogway in the file `onboard/catkin_ws/src/controls/config/oogway.config`.

## Robot testing
Launch:
```bash
roslaunch offboard_comms serial.launch
```
to connect to thrusters.

Ensure that your panel causes the right thrusters to spin at roughly the right speeds. For testing purposes, set thruster speeds to either 10 or 20.

Launch:
```bash
roslaunch controls controls.launch
```
Ensure that a warning is displayed that another node is publishing to `/offboard/thruster_speeds`. This is because `controls.launch` calls `thruster_controls.py` which publishes zeros to `/offboard/thruster_speeds` when controls are disabled (See #441 for how to enable/disable controls).

[^1]: If there is another node that is publishing to `/offboard/thruster_speeds`, then publishing thruster speeds through the GUI will not work due to the overlap. Thus, if another node is publishing to this topic, display a warning message to the user indicating that setting thruster speeds will not work.

viethungle-vt1401:
A Thruster Speeds Panel that allows users to subscribe to the "/offboard/thruster_speeds" topic and read live thruster speed values from the robot. Users can also publish custom thruster speed values to that topic. Closes #451.

Additionally, the shared theme has been updated with dark mode.


foxglove/extensions/thruster-speeds-panel/src/thrusterConfigs.ts
vedarshshah, line 1: Add comments to this file.

vedarshshah (changes requested):
After #484 is merged in, will review all other changed files (those not in `thruster-speeds-panel` dir).

vedarshshah (approved):
LGTM!

PR #484, opened 2023-11-19
Foxglove Custom Messages

### LINKED ISSUE #479: Foxglove Custom Messages
# Background

To publish to topics with a message type part of our `custom_msgs`, Foxglove requires TypeScript definitions for `custom_msgs`. Additionally, as part of publishing, Foxglove requires a map of all datatypes included in the message format; this applies to standard ROS messages, and custom messages. Finally, it would be useful to have TypeScript type definitions for all messages, services, and actions, to enable their static type checking.

Currently, none of these exist. They can be manually created and hardcoded into our extensions, but this would not be good coding practice as it would be difficult to update the extensions when the message definitions change.  An automatic method of generating these, shared across all extensions, would be best. 

# Purpose
Develop automatic methods to generate the following in a way that can be imported into any extension that requires them.
- For each custom message, a `MessageDefinition`.
- For each standard and custom message, a datatypes map.
- For each standard and custom message, service, and action, a type definition.

# Notes
[ros-typescript-generator](https://github.com/Greenroom-Robotics/ros-typescript-generator) can help with generating type definitions.

maxwellmlin:
Closes #479.

All extensions have been updated to import Foxglove custom messages.

Additionally, both the local eslint and our GitHub Actions super-linter have been locked to use eslint v8.53.0.


foxglove/shared/defs/src/messageDefinitions.ts
vedarshshah, line 1: I'll make all necessary changes to this file.
- [ ] Add comments.

foxglove/shared/ros-typescript-generator/src/lib/generateFromRosMsg.spec.ts
vedarshshah, line 1: Can this file be deleted?

foxglove/shared/ros-typescript-generator/src/lib/generateFromRosMsg.ts
vedarshshah, line 1: I'll make the necessary changes to this file.
- [ ] Add comments wherever changes were made from the original file.

foxglove/shared/ros-typescript-generator/README.md
```diff
@@ -0,0 +1,36 @@
+# ROS TypeScript Generator
+A CLI for generating typescript interfaces and enums from ros `.msg`, `.srv`, and `.action` files.
+
+This package has been modified to provide support for the conversion of custom messages without ROS message dependencies. The [original package](https://github.com/Greenroom-Robotics/ros-typescript-generator) was created by Greenroom Robotics.
```
vedarshshah, line 4: Also mention the dependencies that we upgraded to enable the necessary modifications.

foxglove/shared/ros-typescript-generator/src/types/example.d.ts
vedarshshah, line 1: Can this file be deleted?

vedarshshah (changes requested):
- [ ] Does ros-typescript-generator need to be a workspace that is symlinked inside node_modules? Because we use it solely to generate the type definitions that are part of the defs package; we don't import ros-typescript-generator itself in any of our extensions.
- [ ] Make sure the datatype maps have been thoroughly tested. Make sure to test publishing with different datatype maps, including both ROS1 and custom messages.

PR #494, opened 2023-12-14
Nonlinear thrusters allocations

### LINKED ISSUE #324: Integrate Thruster Power Curve & Change Thruster Arduino Comms
# Background

The current controls algorithm assumes thruster force scales linearly with power applied. The actual T200 power curve is provided by Blue Robotics [here](https://bluerobotics.com/store/thrusters/t100-t200-thrusters/t200-thruster-r2-rp/). The controls algorithm should factor in this data to determine the actual 8-bit power value to send to each thruster.

An important note here is that the thrust applied is not symmetric for reverse and forward PWM input values. We need to come up with an algorithm that handles this. This is especially important because some of our thrusters are "reversed", so an expect forward movement actually requires a PWM input in the "reverse" range.

Additionally, the thruster Arduino frequently disconnects as it runs ROS. The communication method needs to be changed so the connection is reliable.

# Purpose
Integrate the nonlinear thruster power curve and improve reliability of connection between robot and thruster Arduino.

# Tasks
- [x] Change controls to publish floats between [-1, 1] to `controls/thruster_allocs` instead of publishing 8-bit integers to `offboard/thruster_speeds`. 
- [x] Develop a node in `offboard_comms` that subscribes to `controls/thruster_allocs` and converts thruster allocs to pulse widths, taking into account the current voltage.
- [x] Modify thruster Arduino code to accept PWMs from the aforementioned node instead of 8-bit integers.

ricohasgithub:
- Added the script `thrusters.cpp` to `offboard_comms`, which subscribes the the voltage sensor and thruster force allocation channels
 - We publish the desired PWM allocation given the current voltage and given thruster force allocation via linear interpolation, whereby we leverage several pre-loaded lookup tables storing information on the relation between voltage, force, and PWM

Closes #324.


onboard/catkin_ws/src/offboard_comms/Arduino Sketchbooks/ThrusterArduino/ThrusterArduino.ino
mbryant2025, line 1: Should remove unused reference to a publisher

onboard/catkin_ws/src/offboard_comms/src/thrusters.cpp
mbryant2025, line 1: Like the implementation here. From my understanding, we are only interpolating between thruster curves to get our PWM. If we really wanted, we could use a greater precision of our voltage (no longer round to 2 places) and interpolate over our individual thruster curves, and then interpolate again between the two curves. Might not be worth it especially if our hardware isn't accurate enough, but just a thought.

mbryant2025, line 1: You define NUM_LOOKUP_ENTRIES in the header file, but do not use it in line 40

mbryant2025 (approved):
Left a few comments about more semantic things, but functionally seems good

onboard/catkin_ws/src/offboard_comms/src/thrusters.cpp
vedarshshah: Just a clarification – we round the _force_ to 2 decimal places, not the voltage; we use full 64-bit floating point precision for the voltage. We lookup the pulse width corresponding to the rounded force in the two thruster power curves for the voltages nearest to the current voltage. Then, we interpolate between the two PWM values using the current voltage, with full `double` precision.

As you indicated, increasing the precision of our force won't help much. Currently, every 0.01 increase in force results in a 2-3 increase in pulse width. So, at most, we could decrease the step size for force to 0.005 to obtain all possible force-to-pulse-width mappings. This is a miniscule difference that won't have much impact on the overall performance of our controls system; there are many other sources of error and inaccuracy that are far greater.

vedarshshah (changes requested):
All the code looks good. I've modified the documentation and comments and did some minor refactoring.

Two things need to happen for approval:
- [ ] Test refactored code in water.
- [ ] @viethungle-vt1401 Update Thruster Speeds Foxglove panel to subscribe/publish to thruster _allocs_ instead of speeds. This will fix the linter and Foxglove issues.

onboard/catkin_ws/src/offboard_comms/Arduino Sketchbooks/ThrusterArduino/ThrusterArduino.ino
```diff
@@ -12,47 +12,77 @@ Adafruit_PWMServoDriver pwm_multiplexer(0x40);
 #define BAUD_RATE 57600
 #define NUM_THRUSTERS 8
 #define THRUSTER_TIMEOUT_MS 500
+#define THRUSTER_STOP_PWM 1500
+#define THRUSTER_PWM_OFFSET 31 // Apply offset of +31 for oogway, remove for cthulhu
```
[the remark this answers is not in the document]
mbryant2025: Finished implementing this functionality in #[edfaa60](https://github.com/DukeRobotics/robosub-ros/pull/494/commits/edfaa60d6ea21ef0aa96234855175c49b54a3ab5) @vedarshshah

onboard/catkin_ws/src/offboard_comms/Arduino Sketchbooks/ThrusterArduino/ThrusterArduino.ino
```diff
@@ -63,6 +65,10 @@ void setup()
         thrusters[i].initialize(&pwm_multiplexer);
         thrusters[i].attach(i);
     }
+
+    // Write the 1500 PWM to all thrusters to stop them
+    for (uint8_t i = 0; i < NUM_THRUSTERS; ++i)
+        thrusters[i].write(THRUSTER_STOP_PWM + THRUSTER_PWM_OFFSET);
```
vedarshshah, line 71: @mbryant2025 I don't think this loop is necessary as `THRUSTER_STOP_PWM + THRUSTER_PWM_OFFSET` will be written to all thrusters in `loop()`.

onboard/catkin_ws/src/offboard_comms/Arduino Sketchbooks/ThrusterArduino/offset.h
vedarshshah, line 1: @mbryant2025 Should we commit this file? Or untrack and gitignore it?

onboard/catkin_ws/src/offboard_comms/scripts/dual_upload.sh
```diff
@@ -63,3 +65,5 @@ if [ "$ARD_UPLOAD" = true ]; then
     arduino-cli upload -b arduino:megaavr:nona4809 -p "$PORT1" "$SRC_CODE1"
     arduino-cli upload -b arduino:megaavr:nona4809 -p "$PORT2" "$SRC_CODE2"
 fi
+
+# onboard/catkin_ws/src/offboard_comms/scripts/copy_offset.sh 0 # Remove the offset file from the compile folder
```
vedarshshah, line 69: @mbryant2025 Should this be uncommented?

vedarshshah (changes requested):
@mbryant2025 See comments below. Also, update the `offboard_comms` README to mention the changes made regarding how the PWM offset is included.

maxwellmlin:
Added a commit that closes #502 for thruster allocs.

vedarshshah (approved):

PR #513, opened 2024-01-27
Pressure Arduino Bug Fix

vedarshshah:
Removed semicolons after `#define` statements that were causing compile errors.


mbryant2025 (approved):
need to add more semicolons

PR #517, opened 2024-01-27
Pressure Fixes

vedarshshah:
Updated pressure Arduino FTDI string in `offboard_comms` package and pressure serial read error message.


mbryant2025 (approved):
Unfortunate that we have the FTDI string in multiple places, will open an issue for that

PR #521, opened 2024-02-04
Controls layouts

maxwellmlin:
Added more layouts to visualize controls.
- `controls-power.json`: Plots the set power scaled, actual power, and their difference for each axis.
- `pid-info.json`: Builds off of `controls-monitor.json`. Additionally, plots the filtered error, filtered derivative, and provided derivative for each axis.


vedarshshah (commented):
LGTM, but hold off on merging for now as pid-info layout may need to be modified based on changes to controls.

vedarshshah (approved):
Layouts are in a good spot now; don't expect any updates anytime soon. @maxwellmlin Please review the recent changes I made before merging. Especially make sure the velocity layout has correct topics and names, as it wasn't tested at the pool.

PR #524, opened 2024-02-15
Offboard Comms

### LINKED ISSUE #518: FTDI string in multiple places
Should have a global config for the Arduino FTDI string so we don't have to remember the 3 places to change it.

vedarshshah:
Variety of updates to `offboard_comms` and `data_pub`. Highlights: resolution of pressure sensor hangup issue, easily maintainable and configurable CLI for uploading to Arduinos, and uniform config file structure.

- Pressure Arduino
  - Updated MS5837 library to latest version from BlueRobotics.
  - Modified MS5837 library to return success/error codes.
  - Added timeout to prevent pressure sensor hangups.
  - Refactored sketch to continue to print voltage despite pressure hangup.
  - Removed unnecessary libraries for Servos and ESCs.
  - Updated measurement of Arduino's true onboard voltage.
- Thruster Arduino
  - Remove unnecessary MS5837 library.
  - Removed unnecessary headers.
- CLI
  - Developed new CLI for installing libraries, finding ports, compiling, and uploading to Arduinos that is easily maintainable and configurable.
  - Removed bash scripts whose functionalities are now covered by the CLI.
  - Added config files (one for each robot) to `offboard_comms` with Arduino configuration.
- Data Pub
  - Replaced old config files with just one file per robot containing DVL FTDI string and axes negation configuration.
  - Refactored `dvl_to_odom.py` to use `transforms3d` instead of `tf.transformations` for compatibility with ROS 2.
  - Refactored `sensors.py` to obtain FTDI string of pressure Arduino from `offboard_comms`.
  - Refactored `sensors.py` to avoid unnecessary errors when starting to read over serial.
- Other
  - Added alias `thruster-test` to spin all thrusters.
  - Updated GitHub Actions to use new CLI instead of `arduino_upload.sh` (now deleted).
  - Updated `offboard_comms` and `data_pub` READMEs so they are up-to-date with the latest changes.
  - Updated `serial.launch` to use new CLI to obtain port of thruster Arduino.
  - Refactored `copy_offset.sh` to use `rospack` to find path to `offboard_comms` package dynamically instead of hardcoding it.
  - Added provided voltage to warning message in `thrusters.cpp` when voltage is out of bounds.

Closes #518.


vedarshshah:
What each person is reviewing:
@maxwellmlin CLI only.
@Will-Denton Updates to data_pub.
@mbryant2025 Everything else.

mbryant2025 (approved):

onboard/catkin_ws/src/offboard_comms/scripts/arduino.py
```diff
@@ -0,0 +1,441 @@
...
+    for arduino_name in arduino_names:
+        try:
+            ports[arduino_name] = get_arduino_port(arduino_name)
+            print(f'{OUTPUT_PREFIX}: {arduino_name.capitalize()}: {ports[arduino_name]}')
+        except StopIteration:
+            error = True
+            print(f'{OUTPUT_PREFIX}: FATAL ERROR: Could not find port of {arduino_name} arduino.')
+
+    if error:
+        print(f'{OUTPUT_PREFIX}: Could not upload sketches to arduinos due to errors in finding ports '
```
maxwellmlin, line 341: Would it be more appropriate to raise an exception here?

onboard/catkin_ws/src/offboard_comms/scripts/arduino.py
```diff
@@ -0,0 +1,441 @@
...
+          f'Please check if the file exists and is in valid YAML format.')
+    exit(1)
+
+# ======================================================================================================================
+# Helper functions
+
+
+def run_command(command: Union[str, Sequence[str]], print_output: bool, env_updates: Optional[dict] = None,
+                path_to_run_at: str = None) -> None:
+    """
```
maxwellmlin, line 63:
```suggestion
                path_to_run_at: Optional[str] = None) -> None:
```

onboard/catkin_ws/src/offboard_comms/scripts/arduino.py
```diff
@@ -0,0 +1,441 @@
...
+            env[key] = str(value)
+
+    print(f'{OUTPUT_PREFIX}: CMD: {command}')
+    subprocess.run(command, shell=True, check=True, stdout=subprocess.DEVNULL if not print_output else None,
+                   stderr=subprocess.STDOUT, env=env, cwd=path_to_run_at)
+
+
+def run_commands(commands: Sequence[str], print_output: bool, env_updates: Optional[dict] = None,
+                 path_to_run_at: str = None) -> None:
+    """
```
maxwellmlin, line 91:
```suggestion
                 path_to_run_at: Optional[str] = None) -> None:
```

maxwellmlin (approved):
Left a few considerations, but LGTM! otherwise

onboard/catkin_ws/src/offboard_comms/scripts/arduino.py
vedarshshah: Yes, especially if this function would be called from another function. However, that is not the case here, and is very unlikely to be the case in the future. The downside with raising an exception is that it prints the traceback, which is not helpful here and would be confusing for anyone not programming-oriented. There also doesn't seem to be a good way to raise an exception in Python without a traceback. Thus, printing the error message and simply returning keeps the output clean and coveys the message without being too confusing.

Will-Denton (approved):
Love to see it!

PR #526, opened 2024-03-01
Sonar Distance Algorithm

Will-Denton:
Sonar dist algorithm:
- findx x,y point reliablly
- thread safe
- finds normal angle


maxwellmlin:
Another senior dev moment

PR #528, opened 2024-03-10
Controls

### LINKED ISSUE #454: Controls Rewrite
# Background

The current controls architecture provides a basic level of control. It provides the robot with the ability to move in a given direction and is roughly accurate relative to state. It has been sufficient to complete the gate and buoy tasks, albeit unreliably.

It has many flaws that limit the stability of the robot and complicate task planning. Therefore, we are rewriting the entirety of controls.

# Purpose

Rewrite the entirety of the robot's controls system, improving its efficiency and stability.

vedarshshah:
Complete rewrite of controls. Highlights include:
- Rewritten entirely in C++
- PID tuning on-the-fly, without needing to restart the system
- Offset the robot's buoyancy
- Support for cascaded and non-cascaded PID
- Quadratic programming for more accurate thrust allocation
- Use different forms of control for each axes at the same time
- Publishes many internal values for tuning, monitoring, and debugging
- Improved efficiency
- Thoroughly documented
- Added C++ linter based on clang-format

Closes #454.


onboard/catkin_ws/src/controls/src/drc_pid.cpp
```diff
@@ -0,0 +1,139 @@
...
+    return (1 / (1 + (c * c) + (1.414 * c))) *
+           (values.at(2) + (2 * values.at(1)) + values.at(0) - ((c * c) - (1.414 * c) + 1) * filtered_values.at(2) -
+            ((-2 * c * c) + 2) * filtered_values.at(1));
+}
+
+double PID::run(double error, double delta_time, PIDInfo &info, double provided_derivative) {
+    // If there are validation errors, an exception is not thrown to maintain continuous operation of controls in the
+    // event of a temporary error. Instead, the function returns 0 and prints an error message.
+
+    // If pid_gains is not valid, return 0 and print error
```
viethungle-vt1401, line 57: why check pid_gains and derivative_type here

onboard/catkin_ws/src/controls/src/controls_utils.cpp
```diff
@@ -0,0 +1,465 @@
...
+        wrench_matrix_file_path = CONTROLS_PACKAGE_PATH + "/" + wrench_matrix_file_path;
+        wrench_matrix_pinv_file_path = CONTROLS_PACKAGE_PATH + "/" + wrench_matrix_pinv_file_path;
+    } catch (const std::exception &e) {
+        ROS_ERROR("Exception: %s", e.what());
+        ROS_ASSERT_MSG(false, "Could not read robot config file. Make sure it is in the correct format. '%s'",
+                       ROBOT_CONFIG_FILE_PATH.c_str());
+    }
+}
+
+void ControlsUtils::update_robot_config(std::function<void(YAML::Node &)> update_function, std::string update_name) {
```
viethungle-vt1401, line 394: split to read/write?

onboard/catkin_ws/src/controls/src/controls.cpp
```diff
@@ -0,0 +1,522 @@
...
+    // Get paths to wrench matrix and its pseudoinverse from robot config file
+    std::string wrench_matrix_file_path;
+    std::string wrench_matrix_pinv_file_path;
+
+    // Get PID configuration parameters from robot config file
+    LoopsMap<AxesMap<double>> loops_axes_control_effort_mins;
+    LoopsMap<AxesMap<double>> loops_axes_control_effort_maxes;
+    LoopsMap<AxesMap<PIDDerivativeTypesEnum>> loops_axes_derivative_types;
+    LoopsMap<AxesMap<double>> loops_axes_error_ramp_rates;
+    LoopsMap<AxesMap<PIDGainsMap>> loops_axes_pid_gains;
```
viethungle-vt1401, line 65: why are some initialized here and some in .h file

onboard/catkin_ws/src/controls/src/controls.cpp
```diff
@@ -0,0 +1,522 @@
...
+                                                custom_msgs::SetStaticPower::Response &res) {
+    tf2::fromMsg(req.static_power, static_power_global);
+
+    // Update static power in robot config file
+    // Throws an exception if file could not be updated successfully; will shut down the node
+    ControlsUtils::update_robot_config_static_power_global(static_power_global);
+
+    res.success = true;
+    res.message =
+        res.success ? "Updated static power successfully." : "Failed to update static power. Static power was invalid.";
```
viethungle-vt1401, line 366: failure message here won't be reached

onboard/catkin_ws/src/controls/src/controls.cpp
```diff
@@ -0,0 +1,522 @@
...
+                                               custom_msgs::SetPowerScaleFactor::Response &res) {
+    power_scale_factor = req.power_scale_factor;
+
+    // Update power scale factor in robot config file
+    // Throws an exception if file could not be updated successfully; will shut down the node
+    ControlsUtils::update_robot_config_power_scale_factor(power_scale_factor);
+
+    res.success = true;
+    res.message = res.success ? "Updated power scale factor successfully."
+                              : "Failed to update power scale factor. Power scale factor was invalid.";
```
viethungle-vt1401, line 381: same here

onboard/catkin_ws/src/controls/src/drc_pid.cpp
viethungle-vt1401, line 57: If `pid_gains` can only be updated through `set_pid_gains`, why don't we validate `pid_gains` there instead of doing that every PID loop iteration?

onboard/catkin_ws/src/controls/src/drc_pid.cpp
```diff
@@ -0,0 +1,139 @@
...
+    integral = ControlsUtils::clip(integral, -integral_clamp, integral_clamp);
+
+    // Update derivatives
+    derivs.at(2) = derivs.at(1);
+    derivs.at(1) = derivs.at(0);
+
+    // Calculate derivative
+    double calculated_derivative = (errors.at(0) - errors.at(1)) / delta_time;
+
+    // Update derivative based on derivative type
```
viethungle-vt1401, line 96: Since `derivative_type` is initialized at node creation (when reading the robot config file), and is not changed by any method, we only need to do `derivative_type` validation once at `controls` node creation or `PID` object creation time and only two cases are needed here.

onboard/catkin_ws/src/controls/src/controls_utils.cpp
```diff
@@ -0,0 +1,465 @@
...
+
+        fout.close();
+    } catch (const std::exception &e) {
+        ROS_ERROR("Exception: %s", e.what());
+        ROS_ASSERT_MSG(false, "Could not update %s in robot config file. Make sure it is in the correct format. '%s'",
+                       update_name.c_str(), ROBOT_CONFIG_FILE_PATH.c_str());
+    }
+}
+
+void ControlsUtils::update_robot_config_pid_gains(const LoopsMap<AxesMap<PIDGainsMap>> &loops_axes_pid_gains,
```
viethungle-vt1401, line 422: In each of these 3 update robot config functions, instead of defining a sub-function and calling `update_robot_config` with `update_function`, can we write separate functions to open and close the robot config file (with mutex handling) instead? The function flow is open function -> specific update logic -> close function.

onboard/catkin_ws/src/controls/src/controls.cpp
viethungle-vt1401, line 366: The failure message won't be used since success is set to True right before.

onboard/catkin_ws/src/controls/src/controls.cpp
```diff
@@ -0,0 +1,522 @@
...
+bool Controls::set_power_scale_factor_callback(custom_msgs::SetPowerScaleFactor::Request &req,
+                                               custom_msgs::SetPowerScaleFactor::Response &res) {
+    power_scale_factor = req.power_scale_factor;
+
+    // Update power scale factor in robot config file
+    // Throws an exception if file could not be updated successfully; will shut down the node
+    ControlsUtils::update_robot_config_power_scale_factor(power_scale_factor);
+
+    res.success = true;
+    res.message = res.success ? "Updated power scale factor successfully."
```
viethungle-vt1401, line 380: The failure message won't be used since success is set to True right before.

onboard/catkin_ws/src/controls/src/controls.cpp
```diff
@@ -0,0 +1,522 @@
...
+    // Store transform buffer
+    this->tf_buffer = std::move(tf_buffer);
+
+    // Initialize desired position to have valid orientation
+    desired_position.orientation.w = 1.0;
+
+    // Use desired position as the default control type for all axes
+    ControlsUtils::populate_axes_map<ControlTypesEnum>(ControlTypesEnum::DESIRED_POSITION, control_types);
+
+    // Get paths to wrench matrix and its pseudoinverse from robot config file
```
viethungle-vt1401, line 56: For the config variables, why are some initialized here while others are initialized in 'controls.h`?

onboard/catkin_ws/src/controls/include/thruster_allocator.h
```diff
@@ -0,0 +1,92 @@
...
+    OsqpEigen::Solver solver;
+
+    /**
+     * @brief Get unconstrained thruster allocations for given set power by returning
+     *  `wrench_pinv * set_power`.
+     *
+     * @param set_power Set power.
+     * @param unconstrained_allocs Unconstrained thruster allocations.
+     */
+    void get_pseudoinverse_solution(const Eigen::VectorXd &set_power, Eigen::VectorXd &unconstrained_allocs);
```
viethungle-vt1401, line 46: Since `get_pseudoinverse_solution` and `get_qp_solution` are only used in one context, we can rename them `get_unconstrained_allocs` and `get_constrained_allocs` to make it clear what they're being used for.

maxwellmlin (approved):
Let's gamble, try merging!

onboard/catkin_ws/src/controls/README.md
```diff
@@ -1,172 +1,930 @@
...
+### Structural Principles
+The `controls` package is designed to be modular and extensible. It is divided into several classes, each of which is responsible for a specific aspect of the system. The classes are designed to be as independent as possible, so that they can be modified or replaced without affecting the rest of the system.
 
-## Testing with Simulation
+The `ControlsUtils` namespace contains a variety of functions that fall into one of the following categories:
+- Functions used by multiple classes
+- Functions that perform tasks that would clutter the classes if they were included in them
 
-To initialize the simulation, follow the instructions in the simulation directory.
+The `Controls` class is the main class of the system. It is responsible for initializing the system, running the main loop, and interfacing with ROS. It uses the other classes to perform the tasks required to control the robot. Its primary functionality is achieved in two functions: `state_callback` and `run`.
```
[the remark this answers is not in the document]
vedarshshah: Added a sentence each to the end of the paragraphs discussing `state_callback` and `run`.

onboard/catkin_ws/src/controls/src/drc_pid.cpp
vedarshshah: Validating `pid_gains` is necessary in `run` if the user initializes the class with the default constructor, in which `pid_gains` is an empty map and the `run` function, if called, would fail. If the user initializes the class with the other constuctor or has set all pid gains though `set_pid_gain`, then the `run` function would succeed.

However, what you are implying is correct – it is inefficient to check the validity of `pid_gains` in every iteration of `run`, especially since it is being run hundreds of times a second. Therefore, to avoid the edge case described above, I initialized all `pid_gains` to zero in the default constructor. Now, all pid gains should exist in `pid_gains` so the `run` function does not have to validate every time.

onboard/catkin_ws/src/controls/src/drc_pid.cpp
vedarshshah: This is technically true. However, it's better to be explicit here about how each derivative type affects the behavior of this function. If another derivative type is added in the future and this code is not updated, then it will be easier to debug as the error will be printed.

onboard/catkin_ws/src/controls/src/controls_utils.cpp
vedarshshah: I agree that this would make the functions easier to understand. However, one of the motivations behind this design was to avoid having to copy-paste the `try`/`catch` block in `update_robot_config`, especially the errors in the `catch`. Your proposal would require copy-pasting the `try`/`catch` block into each update function. I am open to considering an alternative proposal that avoids this issue, but for now, I think the solution I implemented is best. Yes, it takes a bit to understand what's going on, but it's not a huge deal for a future developer to spend a few minutes understanding it.

onboard/catkin_ws/src/controls/src/controls.cpp
vedarshshah: Good question. It is a good practice to initialize variables in the cpp file, not in the header file. I have made this change.

onboard/catkin_ws/src/controls/include/thruster_allocator.h
vedarshshah: The pseudoinverse solution is sometimes used for the unconstrained and constrained allocs, and the qp solution is not always used for the constrained allocs. So, I think it's better to name them based on the method of computation. One can look at the `allocate_thrusters` function to see what they're being used for.

PR #545, opened 2024-04-26
Updated Pressure Arduino Onboard Voltage

vedarshshah:
Allows us to get more accurate readings from the voltage sensor.


maxwellmlin (approved):
Let's gamble, try merging

PR #546, opened 2024-04-26
Sensor Fusion Updates

vedarshshah:
Updated sensor fusion configuration from pool tests.


maxwellmlin:
Included in https://github.com/DukeRobotics/robosub-ros/pull/555.

PR #556, opened 2024-09-02
Fix memory leak in foxglove panels from improper usage of states

achim-k:
This PR fixes a memory leak in your sensor status panel (`@duke-robotics/sensors-status-panel`). The main change is to remove `state` from the dependency array which would cause the `useEffect` hook to be executed every time state changes, causing a memory leak

Note: It's likely that there are more of such memory leaks in other panels, I haven't checked these.


maxwellmlin:
Added to https://github.com/DukeRobotics/robosub-ros2/pull/22.

